PR #22, opened 2022-07-26
Add proposal for Provider Hook

s-sen:
Signed-off-by: ssen <mailssen@gmail.com>

### DIFF
--- a/005-OFEP-provider-hook.md
+++ b/005-OFEP-provider-hook.md
@@ -42,8 +42,7 @@ The Hooks should be executed in specific order.
 
 * before: API, Client, Invocation, Provider
   * Provider does initialization after Invocation
-* after:  Invocation, Provider, Client, API  
-  * Provider runs after Invocation so that it can turn off default behavior if the application chooses to override it
+* after: Provider, Invocation, Client, API  
 * error (if applicable): Provider, Invocation, Client, API
 * finally: Provider, Invocation, Client, API
 
--- a/005-OFEP-provider-hook.md
+++ b/005-OFEP-provider-hook.md
@@ -1,0 +1,49 @@
+## 005-OFEP-provider-hook.md
+
+## State: DRAFTING
+
+As a part of the Provider implementation by the Flag Management Systen, there is a need to have standardized support for hooks specific to the Provider. The 
+Provider Hook(s) should be specific to the provider implementation and implement the various defined stages of the Hook interface as applicable to the 
+specific provider. 
+
+## Assumptions
+
+The Provider Hook is transparent to the application developer. The Provider may choose to document certain default behavior/feature in specific stages of the 
+Provider Hook as relevant to the application and the application developer may be allowed to override the default feature. 
+
+## Example Use Case
+
+Flag Management System X provides the option to add Experimentation as an extension of Feature Flag. The user creates a Feature Flag, runs basic on/off 
+evaluations of the Feature targeting specific audience and then wants to run A/B tests and hence adds the Experimentation aspects to the Feature Flag.
+
+The Provider needs to generate event specific to the Experimentation implementation  on successful flag evaluation. The system wants to use a Provider Hook 
+to do the following for each evaluation  
+1. Run initialization operations (Use the _before_ stage of the Hook)
+2. Run an event publish operation on successful flag evaluation operation (Use the _after_ stage of the Hook)
+3. Provide the option to the application developer to override the event publish operation 
+   1. The application developer may choose not to immediately publish the event on evaluation, but decide to publish it (or not) later based on other 
+      application specific logic
+
+For applications not required to override the default behavior, there is no extra step needed to add any other hooks. The Provider Hook is transparent to 
+the application developer.
+
+## Example Workflow
+
+1. At the time of registering the provider (OpenFeature.setProvider), Provider implementation can register one or more Provider Hook(s)
+2. Provider Hook implements the _before_ and _after_ and any other relevant stages to run the relevant provider logic.
+3. Provider documents the relevant default behavior (e.g. automatic publish of Experimentation event on successful evaluation)
+4. Provider provides an implementation of an Invocation Hook to override this default behavior as a part of the Provider SDK. 
+5. If needed, the application developer uses the invocation hook to override specific default behavior of the provider by using Hook Hints e.g. turn off 
+   default event publish behavior
+
+## Hook Ordering
+
+The Hooks should be executed in specific order.
+
+* before: API, Client, Provider, Invocation
+  * Provider does initialization before Invocation
+* after:  Invocation, Provider, Client, API  
+  * Provider runs after Invocation so that it can turn off default behavior if the application chooses to override it
+* error (if applicable): Provider, Invocation, Client, API
+* finally: Provider, Invocation, Client, API
+
--- a/005-OFEP-provider-hook.md
+++ b/005-OFEP-provider-hook.md
@@ -40,8 +40,8 @@ the application developer.
 
 The Hooks should be executed in specific order.
 
-* before: API, Client, Provider, Invocation
-  * Provider does initialization before Invocation
+* before: API, Client, Invocation, Provider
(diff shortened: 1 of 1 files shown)


005-OFEP-provider-hook.md
```diff
@@ -0,0 +1,49 @@
...
+3. Provider documents the relevant default behavior (e.g. automatic publish of Experimentation event on successful evaluation)
+4. Provider provides an implementation of an Invocation Hook to override this default behavior as a part of the Provider SDK. 
+5. If needed, the application developer uses the invocation hook to override specific default behavior of the provider by using Hook Hints e.g. turn off 
+   default event publish behavior
+
+## Hook Ordering
+
+The Hooks should be executed in specific order.
+
+* before: API, Client, Provider, Invocation
```
beeme1mr, line 43: Could you elaborate on why the provider hooks would run before the invocation hooks? I would have expected the provider hooks to run last in the before hook and first in subsequent hooks. That would allow provider developers to define hooks that are guaranteed to run right before flag evaluation.

beeme1mr (commented):
Thanks for creating this proposal. I think provider hooks would add additional flexibility for provider developers without introducing new concepts.

My only concern is hook ordering. I've elaborated more on this in the proposal itself.

005-OFEP-provider-hook.md
s-sen: In the before stage, the Provider hook can run after the client - I will modify the current order for this.

But for the after stage, one use case is that provider can do some operation that can be overridden by the invocation hook. If the provider hook runs before the invocation hook, the one way the application developer can turn off the default behavior is to use hook hints so that even if the provider hook runs after, it will turn off the default behavior using hook hints. 

What is the use case of the Provider hook to run first in the after stage? If the use case is strong we can mandate that the provider hook can run before the invocation hook - in that case for specific overriding use cases the provider can supply hook hints.

005-OFEP-provider-hook.md
```diff
@@ -0,0 +1,49 @@
...
+3. Provider documents the relevant default behavior (e.g. automatic publish of Experimentation event on successful evaluation)
+4. Provider provides an implementation of an Invocation Hook to override this default behavior as a part of the Provider SDK. 
+5. If needed, the application developer uses the invocation hook to override specific default behavior of the provider by using Hook Hints e.g. turn off 
+   default event publish behavior
+
+## Hook Ordering
+
+The Hooks should be executed in specific order.
+
+* before: API, Client, Invocation, Provider
```
justinabrahms, line 43: I dislike the lack of symmetry here. Would it be reasonable to have providers run their before hooks first? I'd love to maintain a simple "the after is the reverse order of before" if we can.

s-sen: In the 'before' stage, IMO, the Provider should run last. This is because in this stage, the context is mutable and the Provider should be responsible for examining the final form of the context and then run the evaluation.
I am not clear on why the hooks should run the reverse order in the 'after' stage. I am curious about the use cases that rationalizes this.

toddbaert:
Just to clarify: **is there any reason that we couldn't use our existing hooks abstraction for this?** and just attach them at the provider level as specified. My main concern is adding new abstractions... if provider hooks use the same hook abstraction as our existing hooks, I'm more comfortable with this proposal.

I also feel that provider hooks should be "innermost" hooks - they should run last in before, after the application author has run their hooks, and first in error, after, and finally.

005-OFEP-provider-hook.md
toddbaert: > In the 'before' stage, IMO, the Provider should run last. This is because in this stage, the context is mutable and the Provider should be responsible for examining the final form of the context and then run the evaluation.

Agreed.

> I am not clear on why the hooks should run the reverse order in the 'after' stage. I am curious about the use cases that rationalizes this.

Especially for global hooks, having them run in stack order makes a lot of sense - for example, with the OpenTelementry hook, we want to start the span as early as possible and close it at the very end, after all hooks have run, for maximum visibility.

I think in all cases, because these are provider hooks, they should run as close to the provider as possible, so:

```
before: API, Client, Invocation, Provider
after/error/finally: Provider, Invocation, Client, API
```

skyerus: Food for thought, if the ordering of hooks is
```
before: API, Client, Invocation, Provider
after/error/finally: Provider, Invocation, Client, API
```
the only hook that would allow the provider customisability beyond what is already available is the `finally` hook.
A provider already has direct control over what happens immediately before and after a flag call. Whereas there is currently no way for a provider to hook (`finally`) into the execution post success/failure.

toddbaert: >A provider already has direct control over what happens immediately before and after a flag call. Whereas there is currently no way for a provider to hook (finally) into the execution post success/failure.

I think this is good, TBH, that's the business of the application author.

What there currently is _not_, is the ability to easily run arbitrary code before/after flag evaluation in a provider, which can lead to a lot of duplicated code in a provider for each flag resolution type (the context-transformer was a way to avoid this, but it has not been widely adopted and only handled one specific use case).

s-sen: I have modified the after hook order as discussed above. 
I like the idea of running the Provider hook  after the invocation in the 'finally' phase to but this will require more logic in the SDK to ensure the following order for finally: Invocation, Provider, Client, API.

toddbaert:
@s-sen:

>Just to clarify: is there any reason that we couldn't use our existing hooks abstraction for this? and just attach them at the provider level as specified? My main concern is adding new abstractions... if provider hooks use the same hook abstraction as our existing hooks, I'm more comfortable with this proposal.

Can you add any clarification on this? To be clear, I'm just making sure that a provider hook would use the same hook interface that already exists... it seems like that's the case.

toddbaert:
I'd also like to point out I think this makes the `contextTransformer` concept redundant. Since before hooks can transform context, a provider before hook is a great place to do such transformation. I think that if we add provider hooks, we should completely remove the `contextTransformer` concept.

cc @thomaspoignant @beeme1mr @weyert

s-sen:
> Can you add any clarification on this? To be clear, I'm just making sure that a provider hook would use the same hook interface that already exists... it seems like that's the case

Yes, there is no change proposed in the current hook interface. The Provider Hook implementation must use the same interface that is currently defined.

But the application developer should not have to register/add the hook. During running the hooks, the SDK should query the Provider implementation for any provider hooks and if it is provided run the Provider hook in the order that the standard mandates.

toddbaert (approved):
After above discussion, I would support a change to the spec outlining this.

Interested in getting more feedback from others though.

beeme1mr (approved):
Looks good to me.

thomaspoignant:
> 'd also like to point out I think this makes the `contextTransformer` concept redundant. Since before hooks can transform context, a provider before hook is a great place to do such transformation. I think that if we add provider hooks, we should completely remove the `contextTransformer` concept.

100% agree, keeping both will be confusing.

thomaspoignant (approved):

justinabrahms (approved):

justinabrahms:
I really like that this gets rid of context transformers. Thanks, @s-sen!!

toddbaert:
@s-sen I'm going to merge this if you agree.

InTheCloudDan (approved):

beeme1mr:
@s-sen, it looks like we have consensus. Please update the OFEP status to `APPROVED` and I'll merge it in.

Once the OFEP has been merged, feel free to open a PR in the spec repo with the required changes and reference this PR.

Thanks for driving this spec enhancement.

s-sen:
Sorry - I could not do this earlier. I do not need to do any additional changes in the document I assume? 

The next step is to open a PR in the spec repo?

beeme1mr:
@s-sen, nothing else is required here. The next step would be to open a PR in the spec repo and reference this PR.

Thanks for driving this topic. I think it's a great enhancement.

PR #25, opened 2022-07-29
007-OFEP-flag-change-events

toddbaert:
A proposal for subscribing to flag configuration changes.

## UPDATE (Nov 14)

I've made significant changes to this OFEP, taking in to account things discovered in https://github.com/open-feature/js-sdk/pull/316 and https://github.com/open-feature/js-sdk-contrib/pull/142

I also opened a much shorter [OFEP](https://github.com/open-feature/ofep/pull/36) that attempts to address some specific concerns about feature-discovery.

## UPDATE (Feb 2)

I've added a bit more detail here around flag change events and answered some questions! Please take a look.

### DIFF
--- a/007-OFEP-flag-change-events.md
+++ b/007-OFEP-flag-change-events.md
@@ -1,6 +1,6 @@
 ## 007-OFEP-flag-change-events
 
-## State: DRAFTING
+## State: APPROVED
 
 Some flag SDKs support listening for flag value changes or general configuration changes ([launchdarkly](https://docs.launchdarkly.com/sdk/features/flag-changes), [cloudbees](https://docs.cloudbees.com/docs/cloudbees-feature-management/latest/reporting/configuration-fetched-handler), flagd). This can allow us to use an event-based paradigm for consuming flags. Client apps may use feature flags for characteristics that aren't specifically tied to a user-action, making imperative flag evaluation a less-than-ideal solution. Web clients using frameworks such as React need a way to register a callback for when a provider has initialized, given that it's not uncommon for a flag to be evaluated before the provider is ready (during page load). A server application could listen to a flag that changes some operational behavior.
 
--- a/007-OFEP-flag-change-events.md
+++ b/007-OFEP-flag-change-events.md
@@ -21,9 +21,6 @@ How it looks implemented in a Provider:
 ```ts
 import { Provider, EventingProvider, ProviderEvents } from './types';
 
-/**
- * The No-op provider is set by default, and simply always returns the default value.
- */
 class MyEventingProvider implements Provider, EventingProvider {
   readonly metadata = {
     name: 'My Eventing Provider',
--- a/007-OFEP-flag-change-events.md
+++ b/007-OFEP-flag-change-events.md
@@ -12,7 +12,7 @@ Examples:
 
 ## Design
 
-The provider interface and the OpenFeature client would be extended to have new functionality to register handlers for a set of events defined by the SDK (`ProviderEvents`). When a consumer (_application integrator_, _application author_, _integration author_) registers a handler on a client, the client maintains this handler. The provider emits events or runs a callback indicating that it received a certain event, optionally providing data associated with that event. The callbacks registered with the client are then invoked with this data (`EventData`).
+The provider interface and the OpenFeature client would be extended to have new functionality to register handlers for a set of events defined by the SDK (`ProviderEvents`). When a consumer (_application integrator_, _application author_, _integration author_) registers a handler on a client, the client maintains this handler. The provider emits events or runs a callback indicating something happened, optionally providing data associated with that event. The callbacks registered with the client are then invoked with this data (`EventData`).
 
 In the case of the aforementioned flag systems, it is the consumer's responsibility to evaluate any flags in response to the change - flag evaluation would not be automatically performed by the SDK. One reason is that no dynamic context can be reasonably provided in the case of events, since the event is driven by a change in the flag management system, not a user-action. In these cases, the `EventData` would not contain flag values and the application author would have to evaluate flags in the registered handler. This is consistent with the event APIs already existing in those systems.
 
--- a/007-OFEP-flag-change-events.md
+++ b/007-OFEP-flag-change-events.md
@@ -2,7 +2,7 @@
 
 ## State: DRAFTING
 
-Some flag SDKs support listening for flag value changes or general configuration changes ([launchdarkly](https://docs.launchdarkly.com/sdk/features/flag-changes), [cloudbees](https://docs.cloudbees.com/docs/cloudbees-feature-management/latest/reporting/configuration-fetched-handler), flagd). This can allow us to use an event-based paradigm for consuming flags. Client apps may use feature flags for characteristics that aren't specifically tied to a user-action, making imperative flag evaluation a less-than-ideal solution. A server application could listen to a flag that changes some operational behavior.
+Some flag SDKs support listening for flag value changes or general configuration changes ([launchdarkly](https://docs.launchdarkly.com/sdk/features/flag-changes), [cloudbees](https://docs.cloudbees.com/docs/cloudbees-feature-management/latest/reporting/configuration-fetched-handler), flagd). This can allow us to use an event-based paradigm for consuming flags. Client apps may use feature flags for characteristics that aren't specifically tied to a user-action, making imperative flag evaluation a less-than-ideal solution. Web clients using frameworks such as React need a way to register a callback for when a provider has initialized, given that it's not uncommon for a flag to be evaluated before the provider is ready (during page load). A server application could listen to a flag that changes some operational behavior.
 
 Examples:
 
@@ -12,9 +12,9 @@ Examples:
 
 ## Design
 
-The provider interface and the OpenFeature client would be extended to have new functionality to register handlers for a set of events defined by the SDK (`ProviderEvents`). When a _application author_ registers a handler on a client, the client the client maintains this handler. The provider emits events or runs a callback indicating that it received a certain event, optionally providing data associated with that event. The callbacks registered with the client are then invoked with this data (`EventData`).
+The provider interface and the OpenFeature client would be extended to have new functionality to register handlers for a set of events defined by the SDK (`ProviderEvents`). When a consumer (_application integrator_, _application author_) registers a handler on a client, the client maintains this handler. The provider emits events or runs a callback indicating that it received a certain event, optionally providing data associated with that event. The callbacks registered with the client are then invoked with this data (`EventData`).
 
-In the case of the aforementioned flag systems, flags are not actually evaluated when configurations are changed. One reason is that no dynamic context can be reasonably provided in the case of events, since the event is driven by a change in the flag management system, not a user-action. In these cases, the `EventData` would not contain flag values and the application author would have to evaluate flags in the registered handler. This is consistent with the event APIs already existing in those systems.
+In the case of the aforementioned flag systems, it is the consumer's responsibility to evaluate any flags in response to the change - flag evaluation would not be automatically performed by the SDK. One reason is that no dynamic context can be reasonably provided in the case of events, since the event is driven by a change in the flag management system, not a user-action. In these cases, the `EventData` would not contain flag values and the application author would have to evaluate flags in the registered handler. This is consistent with the event APIs already existing in those systems.
 
 How it looks implemented in a Provider:
 
--- a/007-OFEP-flag-change-events.md
+++ b/007-OFEP-flag-change-events.md
@@ -6,59 +6,67 @@ Some flag SDKs support listening for flag value changes or general configuration
 
 Examples:
 
-- A web client's banner color could be updated when an associated flag is changed, without any user action.
-- A web service could subscribe to a flag that controls its log-level.
+- A web client's banner color could be updated when an associated flag is changed, without any user action or a page reload.
+- A service could subscribe to a flag that controls its global log-level.
+- A subsystem could be restarted with the new flag value.
 
 ## Design
 
-The provider interface and the OpenFeature client would be extended to have new functionality to register handlers for a particular flag value or the configuration in general. When an _application author_ registers a handler, the provider would react by appropriately registering a listener on it's SDK, polling it's REST API, etc. If the flag (or configuration in general) is updated, the provider would call the registered handler. The handler might be called with the flag key, or perhaps some other metadata pertaining to the configuration update. **Flag values would not be provided to the handler**, the handler would simply run, indicating the associated configuration had changed. The _application author_ would then perform a flag evaluation for the changed flag. This is consistent with the aforementioned flag systems, which do not evaluate the flags when they've changed. One reason is that no dynamic context can be reasonably provided in the case of events, since the event is driven by a change in the flag management system, not a user-action.
+The provider interface and the OpenFeature client would be extended to have new functionality to register handlers for a set of events defined by the SDK (`ProviderEvents`). When a _application author_ registers a handler on a client, the client the client maintains this handler. The provider emits events or runs a callback indicating that it received a certain event, optionally providing data associated with that event. The callbacks registered with the client are then invoked with this data (`EventData`).
 
-Per-flag design:
+In the case of the aforementioned flag systems, flags are not actually evaluated when configurations are changed. One reason is that no dynamic context can be reasonably provided in the case of events, since the event is driven by a change in the flag management system, not a user-action. In these cases, the `EventData` would not contain flag values and the application author would have to evaluate flags in the registered handler. This is consistent with the event APIs already existing in those systems.
+
+How it looks implemented in a Provider:
 
 ```ts
-const client = OpenFeature.getClient();
+import { Provider, EventingProvider, ProviderEvents } from './types';
 
-// subscribe to changes in a string flag called "hex-color", and evaluate it when it's updated
-client.addHandler("hex-color", async (key) => {
-  console.log(
-    `Got update for ${key}, new value is ${await client.getStringValue(
-      key,
-      "000000"
-    )}`
-  );
-});
+/**
+ * The No-op provider is set by default, and simply always returns the default value.
+ */
+class MyEventingProvider implements Provider, EventingProvider {
+  readonly metadata = {
+    name: 'My Eventing Provider',
+  } as const;
+
+  // ...
+
+  // the SDK listens for events, and fires associated handlers the application-author adds.
+  readonly events = new EventEmitter();
+
+  // pollDataSource a conceptual method specific to this provider that fires a callback if the flag source-of-truth of this provider changes.
+  this.pollDataSource((newFlagData) => this.events.emit(ProviderEvents.ConfigurationChanged, newFlagData))
+
+  // ...
+
 ```
 
-General configuration change design:
+How it looks for an `application author`:
 
 ```ts
+OpenFeature.setProvider(new MyEventingProvider());
+
 const client = OpenFeature.getClient();
 
-// subscribe to changes in the configuration in general, and evaluate flagd it when it's updated
-// this might be easier to implement across vendors, since many don't allow subscription to a particular flag, but general configuration changes
-client.on("configuration-change", async (changeMetadata) => {
-  console.log(
-    `Got update for configuration, new value for hex-color is ${await client.getStringValue(
-      key,
-      "000000"
-    )}`
-  );
+// subscribe to ProviderEvents.ConfigurationChanged events
+client.addHandler(ProviderEvents.ConfigurationChanged, (eventData: EventData | undefined) => {
+
+  // the exact structure of the EventData is not covered in this OFEP
+  if (eventData.changes.myFlag) {
+    onMyFlagChange();
+  }
+  onAnyChange();
 });
 ```
 
 ## Benefits
 
-Ability to use event-based flag evaluation paradigms.
+Ability to use event-based flag evaluation paradigms; attaching event handlers for specific occurrences. This is particularly useful for reacting to configuration changes, provider readiness, and errors. Such event-based models are particularly useful for web frameworks such as React and Angular.
 
 ## Caveats
 
-- Not all providers can reasonably implement this... some SDKs don't support subscriptions or configuration change notifications, for example. Should these providers simply never fire events?
-- Implementation in providers will likely be more divergent than imperative evaluation (based on demo work below).
-- Re-evaluating a flag after the handler fires sometimes means another round-trip, but most of the time in this case, things are locally cached - really dependant on SDK implementation. This is the pattern used by some vendors though, so it's not unprecedented.
-- Some SDKs can't identify individual flags have changed, so all registered handlers must fire (see cloudbees provider demo). This might not be such a big deal since they are likely no-ops.
-- We may need a way to "shutdown" providers, cancelling all handlers.
-- We may need a way to "move" handlers from one provider to another if the provider is changed (recreate them in the new provider).
-- We should probably consider other events, like "SDK-READY" as well, which would fit well into design 2, above.
+- Not all providers can reasonably implement this... some SDKs don't support subscriptions, for example. Should these providers simply never fire events? I've attempted to partially address this with: https://github.com/open-feature/ofep/pull/36
+- We may need a way to "shutdown" providers, cancelling all handlers - this is addressed by: https://github.com/open-feature/ofep/pull/30
 
 ## Demo
 
--- a/007-OFEP-flag-change-events.md
+++ b/007-OFEP-flag-change-events.md
@@ -12,7 +12,7 @@ Examples:
 
 ## Design
 
-The provider interface and the OpenFeature client would be extended to have new functionality to register handlers for a set of events defined by the SDK (`ProviderEvents`). When a consumer (_application integrator_, _application author_) registers a handler on a client, the client maintains this handler. The provider emits events or runs a callback indicating that it received a certain event, optionally providing data associated with that event. The callbacks registered with the client are then invoked with this data (`EventData`).
+The provider interface and the OpenFeature client would be extended to have new functionality to register handlers for a set of events defined by the SDK (`ProviderEvents`). When a consumer (_application integrator_, _application author_, _integration author_) registers a handler on a client, the client maintains this handler. The provider emits events or runs a callback indicating that it received a certain event, optionally providing data associated with that event. The callbacks registered with the client are then invoked with this data (`EventData`).
 
 In the case of the aforementioned flag systems, it is the consumer's responsibility to evaluate any flags in response to the change - flag evaluation would not be automatically performed by the SDK. One reason is that no dynamic context can be reasonably provided in the case of events, since the event is driven by a change in the flag management system, not a user-action. In these cases, the `EventData` would not contain flag values and the application author would have to evaluate flags in the registered handler. This is consistent with the event APIs already existing in those systems.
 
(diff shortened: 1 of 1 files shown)


007-OFEP-flag-change-events.md
```diff
@@ -0,0 +1,66 @@
...
+- A web service could subscribe to a flag that controls its log-level.
+
+## Design
+
+The provider interface and the OpenFeature client would be extended to have new functionality to register handlers for a particular flag value or the configuration in general. When a _application author_ registers a handler, the provider would react by appropriately registering a listener on it's SDK, polling it's REST API, etc. If the flag (or configuration in general) is updated, the provider would call the registered handler. The handler might be called with the flag key, or perhaps some other metadata pertaining to the configuration update. **Flag values would not be provided to the handler**, the handler would simply run, indicating the associated configuration had changed. The _application author_ would then perform a flag evaluation for the changed flag. This is consistent with the aforementioned flag systems, which do not evaluate the flags when they've changed. One reason is that no dynamic context can be reasonably provided in the case of events, since the event is driven by a change in the flag management system, not a user-action.
```
justinabrahms, line 14:
```suggestion
The provider interface and the OpenFeature client would be extended to have new functionality to register handlers for a particular flag value or the configuration in general. When an _application author_ registers a handler, the provider would react by appropriately registering a listener on it's SDK, polling it's REST API, etc. If the flag (or configuration in general) is updated, the provider would call the registered handler. The handler might be called with the flag key, or perhaps some other metadata pertaining to the configuration update. **Flag values would not be provided to the handler**, the handler would simply run, indicating the associated configuration had changed. The _application author_ would then perform a flag evaluation for the changed flag. This is consistent with the aforementioned flag systems, which do not evaluate the flags when they've changed. One reason is that no dynamic context can be reasonably provided in the case of events, since the event is driven by a change in the flag management system, not a user-action.
```

007-OFEP-flag-change-events.md
```diff
@@ -0,0 +1,66 @@
...
+```
+
+General configuration change design:
+
+```ts
+const client = OpenFeature.getClient();
+
+// subscribe to changes in the configuration in general, and evaluate flagd it when it's updated
+// this might be easier to implement across vendors, since many don't allow subscription to a particular flag, but general configuration changes
+client.on("configuration-change", async (changeMetadata) => {
```
justinabrahms, line 39: Instead of having both methods, why don't we have changeMetadata just contain a list of flags which were updated if applicable?

justinabrahms (commented):
I think more broadly, we should have some guidance for SDKs which don't support all OF operations. Like.. if folks don't support doubles or events.. how should they behave.

InTheCloudDan:
> > I think more broadly, we should have some guidance for SDKs which don't support all OF operations. Like.. if folks don't support doubles or events.. how should they behave.
> 
> Agreed. Thoughts on how to accomplish this? Maybe a non-normative section of the spec or something that discusses the general rules around this?

We talked about this some on the community call. I think @beeme1mr was going to follow up with some ideas around naming? One thing that was talked about is that Events like this may not even fit in the `Core` (just current example name) and maybe should be an `Extension` to the spec, because if it's in `Core` every provider should in theory have support for it.

007-OFEP-flag-change-events.md
InTheCloudDan: Do we have an example of `changeMetadata`? I'm only familiar with the LaunchDarkly implementation of flag listeners that can listen for all changes, or a change given a `flag` and a `context`.

toddbaert: > Do we have an example of `changeMetadata`?

I don't at this point, it was just food for thought. I somewhat doubt we could get any consistent information across different SDKs that could be useful there.

[Cloudbees has some metadata about the change](https://docs.cloudbees.com/docs/cloudbees-feature-management/latest/reporting/configuration-fetched-handler), but again, it seems like this would be widely different across SDKs, so we might want to leave out any sort of parameter here.

davejohnston: I like the ```configuration-change``` approach, but is there a half-way house, were we can define an optional filter.   I'm wondering if you had an application connected to a project with a few hundred flags, but only cared about 10, would you get alot of unnecessary interruptions ?

007-OFEP-flag-change-events.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Caveats
+
+- Not all providers can reasonably implement this... some SDKs don't support subscriptions or configuration change notifications, for example. Should these providers simply never fire events?
+- Implementation in providers will likely be more divergent than imperative evaluation (based on demo work below).
+- Re-evaluating a flag after the handler fires sometimes means another round-trip, but most of the time in this case, things are locally cached - really dependant on SDK implementation. This is the pattern used by some vendors though, so it's not unprecedented.
+- Some SDKs can't identify individual flags have changed, so all registered handlers must fire (see cloudbees provider demo). This might not be such a big deal since they are likely no-ops.
+- We may need a way to "shutdown" providers, cancelling all handlers.
+- We may need a way to "move" handlers from one provider to another if the provider is changed (recreate them in the new provider).
+- We should probably consider other events, like "SDK-READY" as well, which would fit well into design 2, above.
```
davejohnston, line 61: I think an event like SDK-READY would make alot of sense.   The current evaluation spec lets us provide a default value, which could be returned during an application startup before the SDK is ready, but there is no way for the consumer to know they got the default vs what they may have wanted.  Being able to listen of SDK-READY effectively lets them handle this, without blocking an application from starting up.

007-OFEP-flag-change-events.md
```diff
@@ -0,0 +1,66 @@
...
+      key,
+      "000000"
+    )}`
+  );
+});
+```
+
+## Benefits
+
+Ability to use event-based flag evaluation paradigms.
```
weyert, line 51: Could use a few more sentences what the `event-based flag evaluation paradigms` are? Is this a common thing along feature flag systems? (first time I hear about it)

007-OFEP-flag-change-events.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Caveats
+
+- Not all providers can reasonably implement this... some SDKs don't support subscriptions or configuration change notifications, for example. Should these providers simply never fire events?
```
weyert, line 55: I think it would make sense for this proposal to answer this question

007-OFEP-flag-change-events.md
```diff
@@ -0,0 +1,66 @@
...
+- Implementation in providers will likely be more divergent than imperative evaluation (based on demo work below).
+- Re-evaluating a flag after the handler fires sometimes means another round-trip, but most of the time in this case, things are locally cached - really dependant on SDK implementation. This is the pattern used by some vendors though, so it's not unprecedented.
+- Some SDKs can't identify individual flags have changed, so all registered handlers must fire (see cloudbees provider demo). This might not be such a big deal since they are likely no-ops.
+- We may need a way to "shutdown" providers, cancelling all handlers.
```
weyert, line 59: I think we could add this to this proposal or raise a separate improvement proposal for this. I think this is important to have -- to ensure resources are getting correctly disposed off

weyert (changes requested):

007-OFEP-flag-change-events.md
```diff
@@ -0,0 +1,66 @@
...
+
+Per-flag design:
+
+```ts
+const client = OpenFeature.getClient();
+
+// subscribe to changes in a string flag called "hex-color", and evaluate it when it's updated
+client.addHandler("hex-color", async (key) => {
```
skyerus, line 22: +1 for this direct approach, looping through a list of changed flags seems unnecessarily cumbersome (other than to support vendors that don't offer this)

james-milligan, line 22: I like the look of this approach over the broad `configuration-change`.

May I also suggest the addition of a `client.addHandlerWithContext` method taking the `EvaluationContext` as an argument. In processes where the `EvaluationContext` is static for the full lifecycle of a process or client instance this will allow for an on change event to a specific flag. This would also reduce the need for `client.get{type}Value` calls at the top of each handler function.

```
client.addHandlerWithContext("hex-color", {"foo":"bar"}, (value) => {
     console.log(`hex-color value change: ${value}`);
})
```

justinabrahms: I don't follow this. What does the context provide here? It doesn't seem to be passed into your resulting function..

toddbaert: I like the idea of providing a context to the handler, but I don't think it will enable us to get the values with a change event, we'd still need to pass the context and call the evaluation in the handler. That's because most SDKs themselves don't allow you to pass the context to their handler, and don't provide flag values when the handler fires:

>The event data consists only of the flag key. It does not contain a flag value, because in server-side SDKs, flags only have values when they are evaluated for a specific set of user properties. (https://docs.launchdarkly.com/sdk/features/flag-changes)

This is LD doc, but it's not unusual. Cloudbees also doesn't provider values with their event.

007-OFEP-flag-change-events.md
toddbaert: I will add more detail. I didn't want to go too far if there wasn't much interest, but it seems like there is.

007-OFEP-flag-change-events.md
toddbaert: I've left this out of my current implementation. For now we won't pass flag values to the handlers either. They will have to be evaluated.

toddbaert:
I've added this [OFEP](https://github.com/open-feature/ofep/pull/36) to address some of the concerns brought up here.

cc @InTheCloudDan @justinabrahms

007-OFEP-flag-change-events.md
beeme1mr: Addressed or at least mitigated by this proposal.
https://github.com/open-feature/ofep/pull/36

007-OFEP-flag-change-events.md
beeme1mr: Addressed by this proposal.
https://github.com/open-feature/ofep/pull/30

007-OFEP-flag-change-events.md
toddbaert: I've made significant updates to this OFEP, addressing many of the concerns in this thread, hopefully.

007-OFEP-flag-change-events.md
toddbaert: I think I've addressed this.

007-OFEP-flag-change-events.md
```diff
@@ -0,0 +1,74 @@
...
+- A service could subscribe to a flag that controls its global log-level.
+- A subsystem could be restarted with the new flag value.
+
+## Design
+
+The provider interface and the OpenFeature client would be extended to have new functionality to register handlers for a set of events defined by the SDK (`ProviderEvents`). When a _application author_ registers a handler on a client, the client the client maintains this handler. The provider emits events or runs a callback indicating that it received a certain event, optionally providing data associated with that event. The callbacks registered with the client are then invoked with this data (`EventData`).
+
+In the case of the aforementioned flag systems, flags are not actually evaluated when configurations are changed. One reason is that no dynamic context can be reasonably provided in the case of events, since the event is driven by a change in the flag management system, not a user-action. In these cases, the `EventData` would not contain flag values and the application author would have to evaluate flags in the registered handler. This is consistent with the event APIs already existing in those systems.
```
justinabrahms, line 17: Counterpoint: in the mobile world, we don't evaluate on the client. "this thing changed" *is* accompanied with values, because we only dispatch on the things that we know about the user (device type and app version).

007-OFEP-flag-change-events.md
```diff
@@ -0,0 +1,74 @@
...
+
+How it looks for an `application author`:
+
+```ts
+OpenFeature.setProvider(new MyEventingProvider());
+
+const client = OpenFeature.getClient();
+
+// subscribe to ProviderEvents.ConfigurationChanged events
+client.addHandler(ProviderEvents.ConfigurationChanged, (eventData: EventData | undefined) => {
```
justinabrahms, line 52: I think I'd like the ability to scope down this handler to only events of a particular key.
```java
logger.setLogLevel(client.getInteger('log_level', 4));
client.onChange('log_level', (i) -> logger.setLogLevel(i));
```

toddbaert: I understand that, and it many cases it should be possible. It's not possible in all cases however, since there are SDKs which only send events indicating a general change occurred, not the flags changed.

In such cases, it could mean "run all handlers" or "run no handlers" or the provider might provide a configuration option in its constructor to choose between these behaviors.

Any takes on this?

beeme1mr: > I understand that, and it many cases it should be possible. It's not possible in all cases however, since there are SDKs which only send events indicating a general change occurred, not the flags changed.

Isn't that where provider features could come into play?

toddbaert: > Isn't that where provider features could come into play?

Provider features don't address the above questions totally. They will give the SDK an idea what this provider supports, but we still have the same fundamental question. Do we fire the handler or not? Do we make it configurable? The handler still has to do or not do _something_.

007-OFEP-flag-change-events.md
moredip, line 17: FWIW, I agree with this approach of pushing responsibility for tracking what needs to be re-evaluated to the consumer.

007-OFEP-flag-change-events.md
moredip: My 2 cents is that we should make implementation of an API like @justinabrahms describes a userland concern, built on top of this core capability, rather than something that's supported directly by the spec/SDKs

jakedoublev:
I'm curious about this proposal given the varying ways flag state changes occur in the vendor SDK's and therefore Providers - polling, streaming, on-fetch, etc. Not to throw a wrench in things, but I'm unsure how best to implement a specific collection of events and their corresponding handlers when we the individual events themselves are so varied across the flagging landscape of vendors/tools. 

It seems like a flag state hook could be created that sets up an event listener within. Pseudocode:
`OpenFeature.addHooks(new EventListenerHook(root)) where "root" is a variable/state node that React or whatever implementation could listen for changes to abstracted away from OpenFeature` or `client.addHooks(new EventListenerHook(someCallbackDefinedByImplementor)) where the implementor passes some state-change logic into the event listener hook constructor`. 

Maybe instead of adding a method to the OpenFeature client and enforcing a strictness on flag vendors/libraries, we could explore development of a library of optional hooks, like the ever-expanding collection of React hooks? 

I think an additional alternative to building out a full event-processing framework within the OpenFeature client would be to look at events in a more black and white context for the moment, considering handling only an event that is analogous to a "breaking change" in Provider/OpenFeature Client state. If shutdown is enabled (see relevant other [issue](https://github.com/open-feature/spec/issues/163) and [PR](https://github.com/open-feature/ofep/pull/30) ), then the implementor of OpenFeature in-app/in-service can shut down the client and reinstantiate if any event is significant enough to require doing so. This seems like a baby step in the right direction.

toddbaert:
> I'm curious about this proposal given the varying ways flag state changes occur in the vendor SDK's and therefore Providers - polling, streaming, on-fetch, etc. Not to throw a wrench in things, but I'm unsure how best to implement a specific collection of events and their corresponding handlers when we the individual events themselves are so varied across the flagging landscape of vendors/tools.

This is the challenge. This proposal attempts to keep things very simple. The only event types defined would be: `PROVIDER_READY`, `PROVIDER_CONFIGURATION_CHANGED`, `PROVIDER_SHUTDOWN` and perhaps `PROVIDER_ERROR`.

The vendor SDKs I've investigated fall into 2 groups with respect to events: those that are capable of emitting some kind of event or firing some callback when the backend configuration is changed, and those that are not. This proposal seems to work effectively for the first category. SDKs in the second category simply wouldn't support events if this proposal was implemented. 

The data received with events (particularly relevant with flag change events) is highly variable, but that's intentionally not addressed by this proposal. If we do attach data to flag change events, it will probably be loosely structured (https://github.com/open-feature/ofep/blob/main/007-OFEP-provider-flag-metadata.md is relevant here) and optional.

> It seems like a flag state hook could be created that sets up an event listener within. Pseudocode: `OpenFeature.addHooks(new EventListenerHook(root)) // "root" is a variable/state node that React or whatever implementation could listen for changes to abstracted away from OpenFeature` or `client.addHooks(new EventListenerHook(someCallbackDefinedByImplementor)) // where the implementor passes some state-change logic into the event listener hook constructor`.

I think this is an interesting idea. Currently hooks only run as part of a evaluation. If we wanted hooks to run with particular events, apart from an evaluation, and have such hooks be the primary means of adding event handlers to OpenFeature objects, I think we'd still have to define some of the interfaces in this proposal. Specifically, we'd still have to define an API that providers could use fire events. The SDK would then run the handlers defined in the hooks. We also might want to extend the hook interface to specifically define a new stage (or stages) for such events.

I think the difference with what you propose is that there wouldn't be more methods on the client, but instead the `application author` would use hooks to attach handlers for events.

> Maybe instead of adding a method to the OpenFeature client and enforcing a strictness on flag vendors/libraries, we could explore development of a library of optional hooks, like the ever-expanding collection of React hooks?

Ya, definitely an interesting proposal. I will think about it a bit more.

> I think an additional alternative to building out a full event-processing framework within the OpenFeature client would be to look at events in a more black and white context for the moment, considering handling only an event that is analogous to a "breaking change" in Provider/OpenFeature Client state. If shutdown is enabled (see relevant other [issue](https://github.com/open-feature/spec/issues/163) and [PR](https://github.com/open-feature/ofep/pull/30) ), then the implementor of OpenFeature in-app/in-service can shut down the client and reinstantiate if any event is significant enough to require doing so. This seems like a baby step in the right direction.

I believe that we need a means of defining a callback that should run when the backend flag configuration changes. Maybe this is another means of doing that, but I'm not 100% sure I understand what you're proposing in this paragraph.

007-OFEP-flag-change-events.md
```diff
@@ -0,0 +1,74 @@
+## 007-OFEP-flag-change-events
+
+## State: DRAFTING
+
+Some flag SDKs support listening for flag value changes or general configuration changes ([launchdarkly](https://docs.launchdarkly.com/sdk/features/flag-changes), [cloudbees](https://docs.cloudbees.com/docs/cloudbees-feature-management/latest/reporting/configuration-fetched-handler), flagd). This can allow us to use an event-based paradigm for consuming flags. Client apps may use feature flags for characteristics that aren't specifically tied to a user-action, making imperative flag evaluation a less-than-ideal solution. A server application could listen to a flag that changes some operational behavior.
```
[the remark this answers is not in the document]
toddbaert: Added a blurb about this.

007-OFEP-flag-change-events.md
```diff
@@ -0,0 +1,74 @@
...
+
+Examples:
+
+- A web client's banner color could be updated when an associated flag is changed, without any user action or a page reload.
+- A service could subscribe to a flag that controls its global log-level.
+- A subsystem could be restarted with the new flag value.
+
+## Design
+
+The provider interface and the OpenFeature client would be extended to have new functionality to register handlers for a set of events defined by the SDK (`ProviderEvents`). When a _application author_ registers a handler on a client, the client the client maintains this handler. The provider emits events or runs a callback indicating that it received a certain event, optionally providing data associated with that event. The callbacks registered with the client are then invoked with this data (`EventData`).
```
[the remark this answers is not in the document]
toddbaert: Good point. I've added `application integrator`. Can you help me understand how an [integration author](https://docs.openfeature.dev/docs/specification/glossary#integration-author) (probably somebody who writes hooks) might use this API?

007-OFEP-flag-change-events.md
```diff
@@ -0,0 +1,74 @@
+## 007-OFEP-flag-change-events
+
+## State: DRAFTING
+
+Some flag SDKs support listening for flag value changes or general configuration changes ([launchdarkly](https://docs.launchdarkly.com/sdk/features/flag-changes), [cloudbees](https://docs.cloudbees.com/docs/cloudbees-feature-management/latest/reporting/configuration-fetched-handler), flagd). This can allow us to use an event-based paradigm for consuming flags. Client apps may use feature flags for characteristics that aren't specifically tied to a user-action, making imperative flag evaluation a less-than-ideal solution. A server application could listen to a flag that changes some operational behavior.
+
+Examples:
+
+- A web client's banner color could be updated when an associated flag is changed, without any user action or a page reload.
+- A service could subscribe to a flag that controls its global log-level.
```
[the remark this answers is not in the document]
moredip: oh yeah, I guess you're right. It's a bad example.

007-OFEP-flag-change-events.md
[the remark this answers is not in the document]
moredip: The example that was top of mind for me was someone who's creating a react-specific feature flagging API that's built on top of Open Feature. I think that fits the persona of Integration Author? This is top of mind because I've been creating [exactly that API](https://github.com/moredip/openfeature-playground/blob/e41e472f45eff9b36a54ceb5ff66e8e1546c77c2/libs/openfeature-hooks/src/lib/useFlag.ts#L24) (a react hook for Open Feature), and need these events in order to make it work in an idiomatic react way.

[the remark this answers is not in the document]
toddbaert: You're absolutely right. I was being myopic thinking of hooks. Will add that persona.

moredip:
FWIW, I've experimented with using @toddbaert's [js-sdk eventing PoC](https://github.com/open-feature/js-sdk/pull/316) to  create a more idiomatic react hooks API for Open Feature - [here's my PoC](https://github.com/moredip/openfeature-playground/blob/experimental-provider-events/libs/openfeature-hooks/src/lib/useFlag.ts#L24). 

Some observations based on that experiment:
- an event for when a provider is ready are pretty much a necessity to build a react-friendly API, because of the fact that client-side code will often try and use the provider before it's ready to provide flags.
- an event for when the provider's configuration changes is very nice if you want the UI to reflect flag changes ASAP.
- Overall, it was super easy to use the proposed API to achieve the two points above 👍 
- At least for React, it would be pretty important to have a `removeHandler` to pair with `addHandler`. React has a tendency to do lots of disposal of objects (component unmounting), and we need some way to clear out handlers during unmount.

toddbaert:
> > > I think more broadly, we should have some guidance for SDKs which don't support all OF operations. Like.. if folks don't support doubles or events.. how should they behave.
> > 
> > 
> > Agreed. Thoughts on how to accomplish this? Maybe a non-normative section of the spec or something that discusses the general rules around this?
> 
> We talked about this some on the community call. I think @beeme1mr was going to follow up with some ideas around naming? One thing that was talked about is that Events like this may not even fit in the `Core` (just current example name) and maybe should be an `Extension` to the spec, because if it's in `Core` every provider should in theory have support for it.

I think https://github.com/open-feature/ofep/blob/main/OFEP-provider-metadata-capability-discovery.md has addressed this.

007-OFEP-flag-change-events.md
```diff
@@ -0,0 +1,74 @@
...
+## Design
+
+The provider interface and the OpenFeature client would be extended to have new functionality to register handlers for a set of events defined by the SDK (`ProviderEvents`). When a _application author_ registers a handler on a client, the client the client maintains this handler. The provider emits events or runs a callback indicating that it received a certain event, optionally providing data associated with that event. The callbacks registered with the client are then invoked with this data (`EventData`).
+
+In the case of the aforementioned flag systems, flags are not actually evaluated when configurations are changed. One reason is that no dynamic context can be reasonably provided in the case of events, since the event is driven by a change in the flag management system, not a user-action. In these cases, the `EventData` would not contain flag values and the application author would have to evaluate flags in the registered handler. This is consistent with the event APIs already existing in those systems.
+
+How it looks implemented in a Provider:
+
+```ts
+import { Provider, EventingProvider, ProviderEvents } from './types';
```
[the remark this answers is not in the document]
toddbaert: I've updated the proposal concerning this ([here is the diff](https://github.com/open-feature/ofep/pull/25/commits/052ceb934830cdeb93042745851958240ab426a6)).

> Do we have a standardized set of events as part of the spec? Or do we keep to just standardizing a few key event names (that would be my preference) and then plan on standardizing other common event typer that might emerge from the community.

I'm interested in this approach, similar to what we have for `Reason`, we can have some predefined ones but allow any string to be used here.

> Also, do we want to support put anything in the spec about any sort of payload that comes with the event?

I've roughly defined the change data structure a bit.

> In your example you're implying that there would be a ConfigurationChanged event that would also include a payload describing flag data

I am proposing metadata, but not "data" (evaluated value). What's being (optionally) provided in the case of a change, is a list of flags changed (so they can be evaluated intelligently) and some free-form metadata (similar to the existing FlagMetadata proposal).

007-OFEP-flag-change-events.md
toddbaert: This is a good point. I think that to maintain as similar an API as possible between client and server impls, it's ideal not to include flag values in the event. In the case of mobile web, as you point out, the flags are probably cached. As long as we consider race conditions, and define our ordering rigorously, I'd hope the cache of flags would be up to date, and subsequent evaluations after the events fire would return accurate flag values.

007-OFEP-flag-change-events.md
toddbaert: @weyert I've also updated the change event payload.

moredip (approved):
Overall I think this is a good proposal. I have some minor questions about whether the events originate from a specific *client* or globally, but I think we can resolve that during the actual spec-ing part.

skyerus (approved):

beeme1mr (approved):

007-OFEP-flag-change-events.md
```diff
@@ -0,0 +1,91 @@
...
+
+Examples:
+
+- A web client's banner color could be updated when an associated flag is changed, without any user action or a page reload.
+- A service could subscribe to a flag that controls its global log-level.
+- A subsystem could be restarted with the new flag value.
+
+## Design
+
+The provider interface and the OpenFeature client would be extended to have new functionality to register handlers for a set of events defined by the SDK (`ProviderEvents`). When a consumer (_application integrator_, _application author_, _integration author_) registers a handler on a client, the client maintains this handler. The provider emits events or runs a callback indicating that it received a certain event, optionally providing data associated with that event. The callbacks registered with the client are then invoked with this data (`EventData`).
```
justinabrahms, line 15: >the client maintains this handler

"client maintains" or "api maintains"? Which level of the hierarchy do these sit?

justinabrahms, line 15:
```suggestion
The provider interface and the OpenFeature client would be extended to have new functionality to register handlers for a set of events defined by the SDK (`ProviderEvents`). When a consumer (_application integrator_, _application author_, _integration author_) registers a handler on a client, the client maintains this handler. The provider emits events or runs a callback indicating something happened, optionally providing data associated with that event. The callbacks registered with the client are then invoked with this data (`EventData`).
```

emits an event when it receives an event sounds really weird.

justinabrahms (approved):

007-OFEP-flag-change-events.md
toddbaert: I've proposed (and PoC'd) this functionality exists on the client.

007-OFEP-flag-change-events.md
```diff
@@ -0,0 +1,91 @@
+## 007-OFEP-flag-change-events
+
+## State: DRAFTING
```
toddbaert, line 3:
```suggestion
## State: APPROVED
```

PR #27, opened 2022-08-06
006-OFEP-flagd-sockets

AlexsJones:
This pull request outlines the proposal to introduce unix sockets into flagD


toddbaert:
I don't have much to say other than I agree with this, even if it's not a top priority, I think it should be done. 

As a possibly interesting data point, redis has some interesting throughput benchmarks for IPC/Unix sockets vs TCP: https://redis.io/docs/reference/optimization/benchmarks/

toddbaert (approved):

beeme1mr (approved):
Please update the OFEP state to APPROVED before merging.

PR #32, opened 2022-08-30
Transaction Context Propagation Proposal

beeme1mr:
Relates to https://github.com/open-feature/spec/issues/81

### DIFF
--- a/007-OFEP-evaluation-context-propagation.md
+++ b/007-OFEP-evaluation-context-propagation.md
@@ -6,7 +6,7 @@ Evaluation context may be used to conditionally control the value returned from 
 
 ## Background
 
-Many languages provide a mechanism for storing variables for the length of a single transaction. This can be used to store evaluation context that can be transparently merged before flag evaluation. The implementation will vary by technology. For example, Java may use [thread local][thread-local], Golang may use [Go Context][go-context], and Node may use [async hooks][async-hooks].
+Many languages provide a mechanism for storing data for the length of a single transaction. This can be used to store evaluation context that can be transparently merged before flag evaluation. The implementation will vary by technology. For example, Java may use [thread local][thread-local], Golang may use [Go Context][go-context], and Node may use [async hooks][async-hooks].
 
 ## Proposal
 
--- a/007-OFEP-evaluation-context-propagation.md
+++ b/007-OFEP-evaluation-context-propagation.md
@@ -14,7 +14,7 @@ I propose that we introduce a way to register a transaction context propagator t
 
 ### Register Context Propagator
 
-In some runtimes (i.e. node), there isn't a native solution for transaction context propagation that would work in all situations. For that reason, it would be beneficial to provide the ability to register a transaction propagator on the global OpenFeature API.
+In some runtimes (e.g. node), there isn't a native solution for transaction context propagation that would work in all situations. For that reason, it would be beneficial to provide the ability to register a transaction propagator on the global OpenFeature API.
 
 ```typescript
 /**
--- a/007-OFEP-evaluation-context-propagation.md
+++ b/007-OFEP-evaluation-context-propagation.md
@@ -2,7 +2,7 @@
 
 ## State: DRAFTING
 
-Evaluation context may be used to conditionally control the value returned from a flag evaluation. You could, for example, enable a feature for all users with a specific email domain. Using evaluation context propagation, application developers can set evaluation context where it's convenient (i.e. an auth service) and have it persist for the length of a transaction (i.e. request or thread).
+Evaluation context may be used to conditionally control the value returned from a flag evaluation. You could, for example, enable a feature for all users with a specific email domain. Using evaluation context propagation, application developers can set evaluation context where it's convenient (e.g. an auth service) and have it persist for the length of a transaction (e.g. a request or thread).
 
 ## Background
 
--- a/007-OFEP-evaluation-context-propagation.md
+++ b/007-OFEP-evaluation-context-propagation.md
@@ -1,0 +1,91 @@
+## NAME
+
+## State: DRAFTING
+
+Evaluation context may be used to conditionally control the value returned from a flag evaluation. You could, for example, enable a feature for all users with a specific email domain. Using evaluation context propagation, application developers can set evaluation context where it's convenient (i.e. an auth service) and have it persist for the length of a transaction (i.e. request or thread).
+
+## Background
+
+Many languages provide a mechanism for storing variables for the length of a single transaction. This can be used to store evaluation context that can be transparently merged before flag evaluation. The implementation will vary by technology. For example, Java may use [thread local][thread-local], Golang may use [Go Context][go-context], and Node may use [async hooks][async-hooks].
+
+## Proposal
+
+I propose that we introduce a way to register a transaction context propagator to the global OpenFeature API. Once registered, evaluation context can be defined at any point in the transaction. When a flag is evaluated, the transaction context is merged based on a defined merging order
+
+### Register Context Propagator
+
+In some runtimes (i.e. node), there isn't a native solution for transaction context propagation that would work in all situations. For that reason, it would be beneficial to provide the ability to register a transaction propagator on the global OpenFeature API.
+
+```typescript
+/**
+ * An example transaction context manager that utilizes async_hooks added in
+ * node 12.17 and marked as stable in node 16.4
+ */
+class AsyncLocalStorageTransactionContext implements TransactionContextManager {
+  private asyncLocalStorage = new AsyncLocalStorage<EvaluationContext>();
+
+  getTransactionContext(): EvaluationContext {
+    return this.asyncLocalStorage.getStore() ?? {};
+  }
+
+  setTransactionContext(
+    context: EvaluationContext,
+    callback: () => void
+  ): void {
+    this.asyncLocalStorage.run(context, callback);
+  }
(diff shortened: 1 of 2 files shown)


007-OFEP-evaluation-context-propagation.md
```diff
@@ -0,0 +1,91 @@
...
+    this.asyncLocalStorage.run(context, callback);
+  }
+}
+
+OpenFeature.setTransactionContextPropagator(
+  new AsyncLocalStorageTransactionContext()
+);
+```
+
+> NOTE: This may not be required for all languages/runtimes
```
toddbaert, line 44: What may not be required? The ability to define the manager?

I suspect we'd always want that, TBH. It seems like that kind of plugability could only be a plus, even if one is registered be default.

beeme1mr: Okay, I'll move this note.

007-OFEP-transaction-context-propagation.md
```diff
@@ -1,18 +1,20 @@
...
 
 ## Background
 
-Many languages provide a mechanism for storing data for the length of a single transaction. This can be used to store evaluation context that can be transparently merged before flag evaluation. The implementation will vary by technology. For example, Java may use [thread local][thread-local], Golang may use [Go Context][go-context], and Node may use [async hooks][async-hooks].
+Many languages provide a mechanism for storing data for the length of a single transaction. This can be used to store transaction context that can be automatically merged with the evaluation context before flag evaluation. The implementation will vary by technology. For example, Java may use [thread local][thread-local], Golang may use [Go Context][go-context], and Node may use [async hooks][async-hooks].
```
evan-bradley, line 11:
```suggestion
Many languages provide a mechanism for storing data for the length of a single transaction. This can be used to store transaction context that can be automatically merged with the evaluation context before flag evaluation. The implementation will vary by technology. For example, Java may use [ThreadLocal][thread-local], Golang may use [Go Context][go-context], and Node may use [async hooks][async-hooks].
```
Nitpick.

007-OFEP-transaction-context-propagation.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Background
+
+Many languages provide a mechanism for storing data for the length of a single transaction. This can be used to store transaction context that can be automatically merged with the evaluation context before flag evaluation. The implementation will vary by technology. For example, Java may use [ThreadLocal][thread-local], Golang may use [Go Context][go-context], and Node may use [async hooks][async-hooks].
+
+## Proposal
+
+This proposal introduces a way to register a transaction context propagator to the global OpenFeature API. Once registered, transaction context can be defined at any point in the transaction. When a flag is evaluated, the transaction context is merged with evaluation context based on the merge order defined below.
```
skyerus, line 15: How does a developer set the transaction context at any point in the transaction?

007-OFEP-transaction-context-propagation.md
```diff
@@ -0,0 +1,91 @@
+## Transaction Context Propagation
+
+## State: DRAFTING
+
+Evaluation context is a container for all contextual data that can be used used to conditionally control the value returned from a flag evaluation. Traditionally, a developer would be responsible for explicitly defining evaluation context during flag evaluation.
```
toddbaert, line 5:
```suggestion
Evaluation context is a container for all contextual data that can be used used to conditionally control the value returned from a flag evaluation. Currently, developers are responsible for explicitly defining and supplying evaluation context during flag evaluation.
```

007-OFEP-transaction-context-propagation.md
```diff
@@ -0,0 +1,91 @@
...
+
+Transaction context, a subset of the evaluation context, is a container for transaction specific data (e.g. user id, user agent, IP) that lives for the duration of a transaction. With transaction context propagation, a developer can set transaction context where it's convenient (e.g. an auth service) and have it persist for the length of a transaction (e.g. a request or thread). OpenFeature will then merge transaction context with evaluation context before flag evaluation.
```
toddbaert, line 7:
```suggestion
_Transaction context_, a subset of the evaluation context, is a container for transaction specific data (e.g. user id, user agent, IP) that lives for the duration of a transaction. With transaction context propagation, a developer can set transaction context where it's convenient (e.g. an auth service) and have it persist for the length of a transaction (e.g. a request or thread). OpenFeature will then merge transaction context with evaluation context before flag evaluation.
```

007-OFEP-transaction-context-propagation.md
toddbaert, line 15:
```suggestion
This proposal introduces a way to register a transaction context propagator to the global OpenFeature API. Once registered, transaction context can be augmented or mutated at any point in the transaction. When a flag is evaluated, the transaction context is merged with evaluation context based on the merge order defined below.
```

toddbaert:
I think this makes a lot of sense, and could really help in applications without a robust dependency injection framework, or some other means of grabbing transaction-scoped data.

For example, in some multi-layered application, information about the subject (authenticated user) may only be available in the web layer. If a flag must be evaluated in some domain logic or data-access layer, it would require the application author to add the relevant user data down the entire call stack to where it's needed; adding an additional param to every called function on a huge function stack... Storing this data in a side-channel, such as thread-local storage or async continuations, allows this information to be accessed anywhere.

toddbaert (approved):
I think this makes a lot of sense, and could really help in applications without a robust dependency injection framework, or some other means of grabbing transaction-scoped data.

For example, in some multi-layered application, information about the subject (authenticated user) may only be available in the web layer. If a flag must be evaluated in some domain logic or data-access layer, it would require the application author to add the relevant user data down the entire call stack to where it's needed; adding an additional param to every called function on a huge function stack... Storing this data in a side-channel, such as thread-local storage or async continuations, allows this information to be passed to flag evaluations without this burden.

I'd like to see a PR adding this to the spec in a platform independent way.

007-OFEP-transaction-context-propagation.md
dyladan:
```suggestion
Flag evaluation may be affected by contextual data including Global, Client, Invocation, and Hook Context.
This proposal defines a new type of context called Evaluation Context.

Evaluation context is a container for all contextual data that can be used used to conditionally control the value returned from a flag evaluation. Traditionally, a developer would be responsible for explicitly defining evaluation context during flag evaluation.
```

007-OFEP-transaction-context-propagation.md
dyladan, line 7:
```suggestion
Flag evaluation may be affected by [evaluation context](link here) data including global, client, invocation, and hook context.
Traditionally, a developer would be responsible for explicitly defining evaluation context during flag evaluation.
This proposal defines a new type of evaluation context called transaction context.

Transaction context is a container for transaction-specific evaluation context (e.g. user id, user agent, IP).
With transaction context propagation, a developer can set transaction context where it's convenient (e.g. an auth service) and it will automatically be applied to all flag evaluations within a transaction.
Transaction context will typically be propagated using a language-specific carrier such as thread local storage or another similar mechanism.
```

007-OFEP-transaction-context-propagation.md
beeme1mr: It would depend on the technology but there's an example using NodeJS later in the proposal.

007-OFEP-transaction-context-propagation.md
```diff
@@ -0,0 +1,94 @@
...
+### Set Transaction Context
+
+Setting transaction context will vary based on the language. In JavaScript, it may look like this:
+
+```typescript
+/**
+ * This example is based on an express middleware.
+ */
+use(req: Request, _res: Response, next: NextFunction) {
+  OpenFeature.setTransactionContext({ targetingKey: req.user.id }, () => {
```
skyerus, line 58: Conceptually I understand that the benefit of using the transaction context here versus the global evaluation context is that this context only lives for the duration of this transaction whereas the global evaluation lives on.
However, I'm not following how this transaction context dies at the end of the transaction?

toddbaert: The transaction context is stored on the [continuation](https://en.wikipedia.org/wiki/Continuation); practically this means that it's accessible only to functions further down the same call stack. This is why the `setTransactionContext()` wraps `next()` (in express, as well as most nodejs web frameworks, `next()` passes the request to the next handler, continuing the execution.

Once the stack is popped back to this point, the data in the continuation is no longer accessible, and GC'd.

beeme1mr:
Hi @morigs, that's a good point. I'll explicitly state that distributed context propagation is out of scope of this OFEP. However, this OFEP is likely a prereq for distributed context propagation using OTel.

By the way, the demo environment includes a proof of concept for distributed context propagation. You can see it [here](https://github.com/open-feature/playground/pull/142).

moredip:
I think this would be really beneficial for some runtimes (i.e. node), and is a really representative example of how the OpenFeature SDK could provide a very useful but fiddly-to-implement capability so that all the various providers don't have to figure it out themselves. 

I would like to see if we could make the API as ergonomic as adding an attribute to an otel span - I think that's a close analog. Likewise it would be great it we could draw inspiration from how otel make the propagator registration as simple as possible for the simple case - maybe you've already done that.

moredip:
@beeme1mr do you think you've gotten enough feedback on this that we could consider it accepted, and move on to speccing it out?

Issue #34, opened 2022-09-08
Android needs


justinabrahms:
Hey folks.

I had a good chat w/ one of eBay's android engineers, @bfasching. He took a look at open feature with an eye towards what would work and what wouldn't.

There were 4 big things that came out of that.

1. We don't have a kotlin provider, so they basically wouldn't use our thing. They need support for `suspend` methods and `flow` to play nicely in their control flow.
2. There's no clarity around blocking vs non-blocking calls. Even though the providers will have a specific implementation, not knowing if you're going to make a blocking call is a big deal in mobile world. The risk to not getting this right is "application not responding" messages (aka freezing the app; eew!)
3. A mobile client should have some awareness of caching, especially in offline use-cases. This is something that can/should be implemented in the client rather than something provider specific.
4. When doing local testing, it's important to support specific overrides. Like, "don't call the backing provider... actually just return 9", even when that may not be the configured default in the code. I *think* we can do that with an aggregated provider that checks a local setup first.

He also demo'd a cool feature we have in our internal SDK that allows conditional defaults based on context. We use that, for instance, to provide different defaults for the German website versus the US website, because the app is aware of which site people use. The specific syntax was kotlin oriented.. but we didn't have an extension point where someone might do something cool with the default value. The best idea I had was we could maybe have another aggregate provider to check to interrogate the context/flag eval options to do the right thing... but that's really hacky.

Interested to hear others' thoughts on the above.

beeme1mr:
Hey @justinabrahms, thanks for putting this together. It's very interesting to learn more about how you're doing mobile feature flagging. I'll have to think a bit more about your specific concerns. They all make sense but I'm not sure the best way to solve them at the moment.

Conditional defaults is an interesting idea. Perhaps we could do something similar if we allowed the after hook to change the return value. I'll have to think about this one a bit more too but I could see how this could be very useful.

justinabrahms:
1 is pretty straight forward, though a fair bit of work.
2 could be solved at the client level by making it explicit that it's blocking or not. If not, we could name the method correctly or do it based on return type?
3 is also do-able, but is probably something that would be a bit complex.

I think conditional defaults may work if we did something like provided a thunk (delayed computation) for calculating the default in addition/instead of the raw value.   So the param would be `OneOf<String | Function<String>>`, as an example.

I'm hoping to have a meeting w/ some iOS folks in the next few weeks.

justinabrahms:
[...]

PR #35, opened 2022-10-21
OFEP: surfacing flag metadata

moredip:
This is still an early draft - opening up a PR for discussion

### DIFF
--- a/00X-OFEP-provider-flag-metadata.md
+++ b/00X-OFEP-provider-flag-metadata.md
@@ -41,7 +41,7 @@ The SDK is responsible for calling `getMetadataForFlag` during every flag evalua
 
 The contents of `FlagMetadata` must be immutable once returned from `getMetadataForFlag`. We won't support adding/removing/editing metadata in hooks, for example.
 
-Format of the metadata attribute is left up to the provider. Maybe we provide suggestions around formatting, and a namespacing prefix (e.g. `"flags-r-us.management-url"). Following otel conventions for span attributes might be smart.
+Format of the metadata attribute is left up to the provider. Maybe we provide suggestions around formatting, and a namespacing prefix (e.g. `"flags-r-us.management-url"`). Following otel conventions for span attributes might be smart.
 
 The SDK will not cache the results of the `getMetadataForFlag` call, and must call it once during every flag evaluation.
 
--- a/00X-OFEP-provider-flag-metadata.md
+++ b/00X-OFEP-provider-flag-metadata.md
@@ -26,7 +26,7 @@ We add an optional `flagMetadata:FlagMetadata` field to the Resolution Details s
 
 The `FlagMetadata` has a type like `Record<string,string|boolean|number>` in typescript, where the key is the metadata attribute and the value is the metadata value.
 
-The flagMetadata field MAY return an empty record.
+The flagMetadata field MAY return an empty record. A missing `flagMetadata` field MUST be interpreted as an empty record.
 
 Since hooks are provided the evaluation context, they would have access to any flag metadata that the provider provides.
 
--- a/00X-OFEP-provider-flag-metadata.md
+++ b/00X-OFEP-provider-flag-metadata.md
@@ -20,31 +20,18 @@ The Open Feature spec will be extended in two ways:
 - a flag provider will be able to provide a bag of metadata attributes for a feature flag. This bag will be a set of key-value pairs, where the key will be a string and the value will be a small set of primitive values.
 - a hook will be able to access these metadata attributes
 
-*note* what follows is a VERY rough strawman proposal for one way to extend the spec. Please poke holes!
+*note* what follows is a strawman proposal for one way to extend the spec. Please poke holes!
 
-### provider interface
-
-We add a `getMetadataForFlag(flagKey:string):FlagMetadata` method to the provider interface. 
+We add an optional `flagMetadata:FlagMetadata` field to the Resolution Details structure - and therefore also to the Evaluation Details structure, since the former is a specified as having a subset of the fields in the latter.
 
 The `FlagMetadata` has a type like `Record<string,string|boolean|number>` in typescript, where the key is the metadata attribute and the value is the metadata value.
 
-If called with an unrecognized `flagKey`, the method MUST return an empty FlagMetadata instance.
+The flagMetadata field MAY return an empty record.
 
-### hook interface
-
-The HookContext interface will gain a `flagMetadata:FlagMetadata` field. 
-
-The SDK is responsible for calling `getMetadataForFlag` during every flag evaluation, and passing whatever was provided into the HookContext.
-
+Since hooks are provided the evaluation context, they would have access to any flag metadata that the provider provides.
 
 ### other details
 
-The contents of `FlagMetadata` must be immutable once returned from `getMetadataForFlag`. We won't support adding/removing/editing metadata in hooks, for example.
+The flagMetadata field should be considered immutable. We won't support adding/removing/editing metadata in hooks, for example.
 
 Format of the metadata attribute is left up to the provider. Maybe we provide suggestions around formatting, and a namespacing prefix (e.g. `"flags-r-us.management-url"`). Following otel conventions for span attributes might be smart.
-
-The SDK will not cache the results of the `getMetadataForFlag` call, and must call it once during every flag evaluation.
-
-## Open Questions
-
-- Should we also expose flag metadata in the `FlagEvaluationDetails`, in case an OF client wants to consume them directly?
--- a/00X-OFEP-provider-flag-metadata.md
+++ b/00X-OFEP-provider-flag-metadata.md
@@ -1,0 +1,50 @@
+## Surfacing flag metadata
+
+## State: DRAFTING
+This proposal lays out a mechanism for flag providers to surface arbitrary flag metadata to Open Feature, and for hooks to access this metadata.
+
+
+## Background
+
+Flag providers maintain metadata about a feature flag, and there are scenarios where a hook might benefit from being able to access this metadata. For example, a flag provider could surface a `management-url` attribute which an open-telemetry hook could then add to an otel span, allowing someone viewing a trace for a feature-flagged operation to easily navigate to a webpage describing that flag in detail.
+
+Different providers will expose different kinds of metadata, and different Application Integrators will want to consume that metadata in different ways. As such, the goal of this proposal is to provide a simple way for the Open Feature runtime to pass generic metadata from a provider to a hook, without Open Feature itself understanding the semantics of that metadata.
+
+## Non-goals
+
+In the future we may wish to standardize the semantics of some key metadata attributes which are common amongst many providers (e.g. `expiration-date`, `owner`, `management-url`) so that provider-agnostic hooks can be created to consume these standard attributes, but that is explicitly NOT in the scope of this OFEP - we would prefer to see which common attributes naturally emerge and then "pave those cow paths".
+
+## Proposal
+
+The Open Feature spec will be extended in two ways:
+- a flag provider will be able to provide a bag of metadata attributes for a feature flag. This bag will be a set of key-value pairs, where the key will be a string and the value will be a small set of primitive values.
+- a hook will be able to access these metadata attributes
+
+*note* what follows is a VERY rough strawman proposal for one way to extend the spec. Please poke holes!
+
+### provider interface
+
+We add a `getMetadataForFlag(flagKey:string):FlagMetadata` method to the provider interface. 
+
+The `FlagMetadata` has a type like `Record<string,string|boolean|number>` in typescript, where the key is the metadata attribute and the value is the metadata value.
+
+If called with an unrecognized `flagKey`, the method MUST return an empty FlagMetadata instance.
+
+### hook interface
+
+The HookContext interface will gain a `flagMetadata:FlagMetadata` field. 
+
+The SDK is responsible for calling `getMetadataForFlag` during every flag evaluation, and passing whatever was provided into the HookContext.
+
+
+### other details
+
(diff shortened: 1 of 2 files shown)


00X-OFEP-provider-flag-metadata.md
```diff
@@ -0,0 +1,50 @@
...
+
+The Open Feature spec will be extended in two ways:
+- a flag provider will be able to provide a bag of metadata attributes for a feature flag. This bag will be a set of key-value pairs, where the key will be a string and the value will be a small set of primitive values.
+- a hook will be able to access these metadata attributes
+
+*note* what follows is a VERY rough strawman proposal for one way to extend the spec. Please poke holes!
+
+### provider interface
+
+We add a `getMetadataForFlag(flagKey:string):FlagMetadata` method to the provider interface.
```
justinabrahms, line 27: I would expect this to come back in the `ProviderEvaluation` object or equivalent?

https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/FeatureProvider.java

00X-OFEP-provider-flag-metadata.md
```diff
@@ -0,0 +1,50 @@
...
+The HookContext interface will gain a `flagMetadata:FlagMetadata` field. 
+
+The SDK is responsible for calling `getMetadataForFlag` during every flag evaluation, and passing whatever was provided into the HookContext.
+
+
+### other details
+
+The contents of `FlagMetadata` must be immutable once returned from `getMetadataForFlag`. We won't support adding/removing/editing metadata in hooks, for example.
+
+Format of the metadata attribute is left up to the provider. Maybe we provide suggestions around formatting, and a namespacing prefix (e.g. `"flags-r-us.management-url"). Following otel conventions for span attributes might be smart.
```
justinabrahms, line 44:
```suggestion
Format of the metadata attribute is left up to the provider. Maybe we provide suggestions around formatting, and a namespacing prefix (e.g. `"flags-r-us.management-url"`). Following otel conventions for span attributes might be smart.
```

00X-OFEP-provider-flag-metadata.md
```diff
@@ -0,0 +1,50 @@
...
+
+The SDK will not cache the results of the `getMetadataForFlag` call, and must call it once during every flag evaluation.
+
+## Open Questions
+
+- Should we also expose flag metadata in the `FlagEvaluationDetails`, in case an OF client wants to consume them directly?
```
justinabrahms, line 50: Seems not very "vendor agnostic"-y so I'd probably call that a code smell. That said, I'm sure there are also reasons why you'd want to. Maybe wait for someone to have such a reason then we add the feature?

00X-OFEP-provider-flag-metadata.md
moredip: Wow, not sure why it didn't occur to me to put it there instead - that makes much more sense!

00X-OFEP-provider-flag-metadata.md
moredip: The only reason I can think of is for an Application Implementor who isn't quite ready to learn how to implement a custom hook, but wants access to the metadata for some sort of internal framework.

00X-OFEP-provider-flag-metadata.md
moredip: @justinabrahms updated the proposal per your suggestion: ace0ba9abe534ec064076749078353e6f791a310

00X-OFEP-provider-flag-metadata.md
moredip: this is a moot point now that I switched the proposal to include the metadata in the Resolution Details - it's now unavoidably exposed to the client.

00X-OFEP-provider-flag-metadata.md
```diff
@@ -0,0 +1,37 @@
...
+
+The Open Feature spec will be extended in two ways:
+- a flag provider will be able to provide a bag of metadata attributes for a feature flag. This bag will be a set of key-value pairs, where the key will be a string and the value will be a small set of primitive values.
+- a hook will be able to access these metadata attributes
+
+*note* what follows is a strawman proposal for one way to extend the spec. Please poke holes!
+
+We add an optional `flagMetadata:FlagMetadata` field to the Resolution Details structure - and therefore also to the Evaluation Details structure, since the former is a specified as having a subset of the fields in the latter.
+
+The `FlagMetadata` has a type like `Record<string,string|boolean|number>` in typescript, where the key is the metadata attribute and the value is the metadata value.
```
justinabrahms, line 27: If you want to make it return a mapping of string to some value type.. then you probably want to just make it a `Structure` for consistency? https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/Structure.java

00X-OFEP-provider-flag-metadata.md
```diff
@@ -0,0 +1,37 @@
...
+Different providers will expose different kinds of metadata, and different Application Integrators will want to consume that metadata in different ways. As such, the goal of this proposal is to provide a simple way for the Open Feature runtime to pass generic metadata from a provider to a hook, without Open Feature itself understanding the semantics of that metadata.
+
+## Non-goals
+
+In the future we may wish to standardize the semantics of some key metadata attributes which are common amongst many providers (e.g. `expiration-date`, `owner`, `management-url`) so that provider-agnostic hooks can be created to consume these standard attributes, but that is explicitly NOT in the scope of this OFEP - we would prefer to see which common attributes naturally emerge and then "pave those cow paths".
+
+## Proposal
+
+The Open Feature spec will be extended in two ways:
+- a flag provider will be able to provide a bag of metadata attributes for a feature flag. This bag will be a set of key-value pairs, where the key will be a string and the value will be a small set of primitive values.
```
agentgonzo, line 20: +1 on the 'may' rather than 'must'. In the simple cases, we don't want to force providers to return an empty bag. The OF SDK should default to an empty bag if the provider does nothing.

00X-OFEP-provider-flag-metadata.md
moredip: Personally I'm sort of reluctant to make it so general that it would also support lists, objects, etc. Feels like it's opening the door to some perverse use cases. 

My 2c is that it'd be better to keep a pretty hard constraint, initially, and then if some valid use cases for richer values come along we could relax that constraint in future spec revisions.

00X-OFEP-provider-flag-metadata.md
```diff
@@ -0,0 +1,37 @@
+## Surfacing flag metadata
+
+## State: DRAFTING
+This proposal lays out a mechanism for flag providers to surface arbitrary flag metadata to Open Feature, and for hooks to access this metadata.
+
+
+## Background
+
+Flag providers maintain metadata about a feature flag, and there are scenarios where a hook might benefit from being able to access this metadata. For example, a flag provider could surface a `management-url` attribute which an open-telemetry hook could then add to an otel span, allowing someone viewing a trace for a feature-flagged operation to easily navigate to a webpage describing that flag in detail.
```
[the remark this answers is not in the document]
moredip: > This use case would appear to me as more of a global than per flag use case. Where you join the `management-url` of the `global` config with the specific flag key to add to otel span.
> 
> Are there additional use case examples that can be added for individual metadata on a flag alone?

I would have though the management url would be per-flag: `https://app.flagrsus.com/flags/1412351` or whatever. I would expect a provider would want it to be an opaque url rather than having each hook build the URL itself.

Other use cases off the top of my head:
- owner: which team or person owns this flag, so I can go ping them and ask whether their flag could be related to some operational issue
- expiration date: should this flag have been retired by now? I've seen teams log warning or even crash hard if their code is referencing stale flags. "Feature flag debt" is a very real thing.
- subsystem: which subsystem is supposed to be controlled by this flag, so I can enforce rules regarding which flags are used where. For example, I might want to enforce that client-side code isn't using a server-side flag, or web-app code isn't using a mobile flag, or whatever.

00X-OFEP-provider-flag-metadata.md
moredip: > I wonder if it is worth making preparations for secure/sensitive and non-secure/benign meta data? Something like an evaluation timestamp might be treated in a different way to a management or internal system URL. My worry would be that sensitive metadata then starts getting logged/surfaced without the knowledge or explicit approval of the developer.
> 
> I've not used OpenTelemetry - does it handle this sort of thing or not?

AFAIK otel doesn't support this feature, and IIRC it's actually causes some security issues with baggage that is accidentally propagated outside of internal otel tooling into 3rd party systems. For example, you call into a SaaS provider's APIs, and you get back a response with baggage describing the internal state of their system 😬 

However, I still feel that this is really more the responsibility of the provider to decide whether it should send sensitive metadata to the client or not. For client-side code anything you put on the wire is going to be available regardless, so seems to me that as a provider you'd be forced to just never put that metadata on the wire.

moredip:
> I wonder if it is worth making preparations for secure/sensitive and non-secure/benign meta data? Something like an evaluation timestamp might be treated in a different way to a management or internal system URL. My worry would be that sensitive metadata then starts getting logged/surfaced without the knowledge or explicit approval of the developer.
> 
> I've not used OpenTelemetry - does it handle this sort of thing or not?

AFAIK otel doesn't support this feature, and IIRC it's actually causes some security issues with baggage that is accidentally propagated outside of internal otel tooling into 3rd party systems. For example, you call into a SaaS provider's APIs, and you get back a response with baggage describing the internal state of their system 😬 

However, I still feel that this is really more the responsibility of the provider to decide whether it should send sensitive metadata to the client or not. For client-side code anything you put on the wire is going to be available regardless, so seems to me that as a provider you'd be forced to just never put that metadata on the wire.

00X-OFEP-provider-flag-metadata.md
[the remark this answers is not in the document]
justinabrahms: Another example: https://github.com/open-feature/ofep/issues/34

> As it relates to flag lifecycle & experimentation, they have a pattern where sometimes they'll query for a flag value early and actually use it much later. In that time, the values may have swapped. As such, I think they want the ability to include other metadata along with the EvaluationDetails (like timestamp of fetch, etc)

00X-OFEP-provider-flag-metadata.md
```diff
@@ -0,0 +1,37 @@
...
+
+The flagMetadata field MAY return an empty record.
+
+Since hooks are provided the evaluation context, they would have access to any flag metadata that the provider provides.
+
+### other details
+
+The flagMetadata field should be considered immutable. We won't support adding/removing/editing metadata in hooks, for example.
+
+Format of the metadata attribute is left up to the provider. Maybe we provide suggestions around formatting, and a namespacing prefix (e.g. `"flags-r-us.management-url"`). Following otel conventions for span attributes might be smart.
```
beeme1mr, line 37: Following the OTel attribute naming rules as a recommendation makes sense to me. You can see what they came up with [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/attribute-naming.md).

beeme1mr (commented):
I like this proposal. @toddbaert could you please take a look a provide feedback?

00X-OFEP-provider-flag-metadata.md
beeme1mr: I think I would prefer primitive types for now too.

00X-OFEP-provider-flag-metadata.md
moredip: ea68f0b makes this clearer

moredip:
> @moredip, would you be able to resolve the conversations that have been addressed? I would like to address any outstanding concerns and try and get this merged soon.

@beeme1mr I marked all conversations here as resolved. LMK if there's anything else you need me to do from my end to move this forward.

00X-OFEP-provider-flag-metadata.md
```diff
@@ -0,0 +1,37 @@
...
+
+Different providers will expose different kinds of metadata, and different Application Integrators will want to consume that metadata in different ways. As such, the goal of this proposal is to provide a simple way for the Open Feature runtime to pass generic metadata from a provider to a hook, without Open Feature itself understanding the semantics of that metadata.
+
+## Non-goals
+
+In the future we may wish to standardize the semantics of some key metadata attributes which are common amongst many providers (e.g. `expiration-date`, `owner`, `management-url`) so that provider-agnostic hooks can be created to consume these standard attributes, but that is explicitly NOT in the scope of this OFEP - we would prefer to see which common attributes naturally emerge and then "pave those cow paths".
```
toddbaert, line 15: I think this is a good thing to mention, and I agree it can (and should) be left out of this OFEP.

00X-OFEP-provider-flag-metadata.md
```diff
@@ -0,0 +1,37 @@
...
+
+The flagMetadata field MAY return an empty record. A missing `flagMetadata` field MUST be interpreted as an empty record.
+
+Since hooks are provided the evaluation context, they would have access to any flag metadata that the provider provides.
+
+### other details
+
+The flagMetadata field should be considered immutable. We won't support adding/removing/editing metadata in hooks, for example.
+
+Format of the metadata attribute is left up to the provider. Maybe we provide suggestions around formatting, and a namespacing prefix (e.g. `"flags-r-us.management-url"`). Following otel conventions for span attributes might be smart.
```
toddbaert, line 37: Only tangentially related, but I move to make `flags-r-us` our de-facto fake flag management system... like our `Contoso`.

toddbaert:
I really like the proposal as well. I think the same flag metadata might also be optionally provided in events, such as `ConfigurationChanged` events mentioned in the [eventing OFEP](https://github.com/open-feature/ofep/pull/25) as well.

toddbaert (approved):

beeme1mr (approved):

PR #36, opened 2022-11-14
feat: provider metadata feature discovery

toddbaert:
Mostly brought up based on comments in the [events OFEP](https://github.com/open-feature/ofep/pull/25), this OFEP proposes a solution for "feature discovery" in the context of providers, in order to signal to the SDK and `application authors` what features are available in a particular provider.

I think this would work well with badges and documentation that consistently identify features and functionality.

### DIFF
--- a/??-OFEP-provider-metadata-feature-discovery.md
+++ b/??-OFEP-provider-metadata-feature-discovery.md
@@ -1,0 +1,36 @@
+## ??-OFEP-provider-metadata-feature-discovery
+
+## State: DRAFTING
+
+This OFEP proposes a solution for "feature discovery" in the context of providers, in order to signal to the SDK and `application authors` what features are available in a particular provider.
+
+## Background
+
+As the OpenFeature specification evolves, it's to be expected that some components may not be able to support certain features. For instance, some providers may not support flag change events (as described in https://github.com/open-feature/ofep/pull/25). 
+
+## Proposal
+
+This OFEP proposes that the [`provider metadata`](https://docs.openfeature.dev/docs/specification/sections/providers#requirement-211) be extended to include optional properties that denote which features are available on the implementing provider. The SDK can then make intelligent decisions and log warnings if features that are not supported by the provider in question, are used. For example, if an `application author` adds an event handler but the registered provider doesn't support events, the SDk can log a warning.
+
+### Example Implementation
+
+```typescript
+
+import { Provider, Functions } from '@openfeature/js-sdk';
+
+/**
+ * The following provider supports `SomeFeature`, `ProviderEvents`, and `SomeOtherFeature`, which are defined by the SDK.
+ */
+class SomeFeatureProvider implements Provider {
+  readonly metadata = {
+    name: 'Some Feature Provider',
+		features: {
+			[Functions.SomeFeature]: true,
+			[Functions.ProviderEvents]: true,
+			[Functions.SomeOtherFeature]: false,
+		}
+  };
+
+	// ...implementation...
+}
+```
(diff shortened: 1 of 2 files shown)


??-OFEP-provider-metadata-feature-discovery.md
```diff
@@ -0,0 +1,36 @@
...
+
+This OFEP proposes a solution for "feature discovery" in the context of providers, in order to signal to the SDK and `application authors` what features are available in a particular provider.
+
+## Background
+
+As the OpenFeature specification evolves, it's to be expected that some components may not be able to support certain features. For instance, some providers may not support flag change events (as described in https://github.com/open-feature/ofep/pull/25). 
+
+## Proposal
+
+This OFEP proposes that the [`provider metadata`](https://docs.openfeature.dev/docs/specification/sections/providers#requirement-211) be extended to include optional properties that denote which features are available on the implementing provider. The SDK can then make intelligent decisions and log warnings if features that are not supported by the provider in question, are used. For example, if an `application author` adds an event handler but the registered provider doesn't support events, the SDk can log a warning.
```
beeme1mr, line 13:
```suggestion
This OFEP proposes that the [`provider metadata`](https://docs.openfeature.dev/docs/specification/sections/providers#requirement-211) be extended to include optional properties that denote which features are available on the implementing provider. The SDK can then make intelligent decisions and log warnings if features that are not supported by the provider in question, are used. For example, if an `application author` adds an event handler but the registered provider doesn't support events, the SDK can log a warning.
```

??-OFEP-provider-metadata-feature-discovery.md
```diff
@@ -0,0 +1,36 @@
...
+
+import { Provider, Functions } from '@openfeature/js-sdk';
+
+/**
+ * The following provider supports `SomeFeature`, `ProviderEvents`, and `SomeOtherFeature`, which are defined by the SDK.
+ */
+class SomeFeatureProvider implements Provider {
+  readonly metadata = {
+    name: 'Some Feature Provider',
+		features: {
```
beeme1mr, line 27: There seems to be an indentation issue here.

??-OFEP-provider-metadata-feature-discovery.md
```diff
@@ -0,0 +1,36 @@
+## ??-OFEP-provider-metadata-feature-discovery
```
beeme1mr, line 1: This highlights an issue with the way the OFEPs are structured. A quick win may be to omit the number from the file name and title.

beeme1mr (commented):
Could you please add a section that talks about backwards compatibility? I would expect new features to be optional and require the provider developer to explicitly add support for the feature.

toddbaert:
> Could you please add a section that talks about backwards compatibility? I would expect new features to be optional and require the provider developer to explicitly add support for the feature.

I think that's covered by the phrasing:

> ...be extended to include optional properties that denote which features are available on the implementing provider...

Since all these are optional, we have backwards compatibility.

@beeme1mr

justinabrahms (approved):

beeme1mr (approved):
I think this proposal addresses a number of concerns we had regarding advanced features that can't be uniformly addressed across all providers.

kinyoklion (approved):

OFEP-provider-metadata-feature-discovery.md
```diff
@@ -0,0 +1,36 @@
+## OFEP-provider-metadata-feature-discovery
+
+## State: DRAFTING
+
+This OFEP proposes a solution for "feature discovery" in the context of providers, in order to signal to the SDK and `application authors` what features are available in a particular provider.
```
moredip, line 5: "feature" is quite an overloaded word in the domain of open feature. For the sake of sanity, could we use "capability" instead? We'd use capability discovery to discover which capabilities a given provider has. Language like "does this provider support the provider events capability" seems pretty natural to me.

toddbaert: Great point, love it.

PR #41, opened 2022-12-23
Add "static-context" paradigm OFEP

toddbaert:
As discussed in the 22nd community meeting, this draft outlines a new "static-context" paradigm to better support client/web use-cases.

The key changes in this paradigm are:

- the lack a context parameter on evaluation methods
- the lack of a context setter on the client class
- a optional method in providers that is called when the global context is changed, which the provider can use to "reconcile" it's context

For more information, please see the OFEP.  

I believe this, combined with [disposal](https://github.com/open-feature/ofep/pull/30) and [events](https://github.com/open-feature/ofep/pull/25), represent the basis for proper client support in the spec.

### DIFF
--- a/OFEP-single-context-paradigm.md
+++ b/OFEP-single-context-paradigm.md
@@ -1,0 +1,84 @@
+## OFEP-single-context-paradigm
+
+## State: DRAFTING
+
+This draft outlines a "single-context" paradigm - an alternative pattern and set of APIs supporting client use-cases.
+
+## Background
+
+In contrast with server-side or other service-type applications, client side apps typically operate in the context of a single user.
+Most feature flagging for these applications have been designed with this in mind.
+In summary, most client/web SDKs operate something like this:
+
+- an initialization occurs, which fetches evaluated flags in bulk for a given context (user)
+- the evaluated flags are cached in the SDK
+- flag evaluations take place against this cache, without a need to provide context (context was already used to evaluate flags in bulk)
+- Functions/methods are exposed on the SDK that signal the cache is no longer valid, and must be reconciled based on a context change. This frequently involves a network request or I/O operation.
+
+This paradigm doesn't fit well with the existing SDK, which, like most server-side SDKs, emphasizes realtime evaluation and context-per-evaluation.
+
+Though not all client side SDKs function in this way, those that do allow context per evaluation can conform to this model fairly easily.
+
+## Design
+
+To better support the "single-context" paradigm we need to define some additional handlers and add some flexibility to current APIs.
+In short, the application author now sets context globally using the existing global context mutator on the OpenFeature API object.
+This single context is used by providers for initialization and fetching evaluated flag values in bulk.
+All evaluations functions are "context-less".
+When the context is changed, providers are signalled to update their cache of evaluated flags.
+
+### Provider changes
+
+#### On-context-set handler
+
+Providers must have a mechanism to understand when their cache of evaluated flags must be invalidated or updated. An `on-context-set` handler can be defined which performs whatever operations are needed to reconcile the evaluated flags with the new context. The OpenFeature SDK calls this handler when the global context is modified.
+
+```typescript
+
+interface Provider {
+  //...
+
+  // a handler called by the SDK when context is modified
+  onContextChange?(oldContext: EvaluationContext, newContext: EvaluationContext): Promise<void>
+
+  //...
+}
+```
+
+#### Remove evaluation-context parameter on resolvers
+
+With all evaluation context represented in the global context, resolvers no longer need it as a parameter. Providers that require context at evaluation can simply access the global evaluation context for use in their evaluation.
+
+```typescript
+
+interface Provider {
+  //...
+  
+  // context parameter is removed from resolver
+  resolveBooleanEvaluation(flagKey: string, defaultValue: boolean): ResolutionDetails<boolean>;
+
+  //...
+}
+```
+
+### Client changes
+
+#### Remove evaluation-context parameter on evaluation functions
+
+Similarly to the changes in the provider, the evaluator functions on the client do not accept a context.
+
+```typescript
+interface Client {
+  //...
+
+  // context parameter is removed from evaluation API
+  getBooleanValue(flagKey: string, defaultValue: boolean, options?: FlagEvaluationOptions): boolean;
+  getBooleanDetails(flagKey: string, defaultValue: boolean, options?: FlagEvaluationOptions): EvaluationDetails<boolean>;
+
+  //...
+}
+```
+
+#### Remove context mutator
+
+The global context has a one-to-one correspondence to the providers cache of evaluated flags. It's unreasonable or impossible to reconcile multiple client-level contexts with this state. The context mutator of the client must be removed.
--- a/OFEP-single-context-paradigm.md
+++ b/OFEP-single-context-paradigm.md
@@ -7,7 +7,7 @@ This draft outlines a "single-context" paradigm - an alternative pattern and set
 ## Background
 
 In contrast with server-side or other service-type applications, client side apps typically operate in the context of a single user.
-Most feature flagging for these applications have been designed with this in mind.
+Most feature flagging SDKs for these applications have been designed with this in mind.
 In summary, most client/web SDKs operate something like this:
 
 - an initialization occurs, which fetches evaluated flags in bulk for a given context (user)
--- a/OFEP-single-context-paradigm.md
+++ b/OFEP-single-context-paradigm.md
@@ -47,21 +47,25 @@ interface Provider {
 
 While the `on-context-set` handler is executing, the cache of resolved flags may be considered "stale". `Provider authors` and `application authors` should understand the consequences of evaluating flags in this state.
 
-#### Remove evaluation-context parameter on resolvers
+#### Initialize function
 
-With all evaluation context represented in the global context, resolvers no longer need it as a parameter. Providers that require context at evaluation can simply access the global evaluation context for use in their evaluation.
+Providers may need access to the static context when they start up.
+Passing this in the provider constructor is not always possible, and ergonomics are improved by separating configuration and evaluation.
 
-```typescript
+An `initialize` function can be optionally implemented by a provider, which defines an parameter for the static context.
 
 interface Provider {
   //...
-  
-  // context parameter is removed from resolver
-  resolveBooleanEvaluation(flagKey: string, defaultValue: boolean): ResolutionDetails<boolean>;
+
+  // a function called by the SDK when the provider becomes active
+  initialize?(context: EvaluationContext): Promise<void>
 
   //...
 }
 ```
+
+> NOTE: The provider interface will retain the context parameter.
+The parameter will be supplied by the SDK from the global context.
 
 ### Client changes
 
--- a/OFEP-single-context-paradigm.md
+++ b/OFEP-single-context-paradigm.md
@@ -1,6 +1,6 @@
 ## OFEP-single-context-paradigm
 
-## State: DRAFTING
+## State: APPROVED
 
 This draft outlines a "single-context" paradigm - an alternative pattern and set of APIs supporting client use-cases.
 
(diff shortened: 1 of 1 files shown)


OFEP-single-context-paradigm.md
```diff
@@ -0,0 +1,84 @@
...
+In short, the application author now sets context globally using the existing global context mutator on the OpenFeature API object.
+This single context is used by providers for initialization and fetching evaluated flag values in bulk.
+All evaluations functions are "context-less".
+When the context is changed, providers are signalled to update their cache of evaluated flags.
+
+### Provider changes
+
+#### On-context-set handler
+
+Providers must have a mechanism to understand when their cache of evaluated flags must be invalidated or updated. An `on-context-set` handler can be defined which performs whatever operations are needed to reconcile the evaluated flags with the new context. The OpenFeature SDK calls this handler when the global context is modified.
```
beeme1mr, line 34:
```suggestion
Providers may need a mechanism to understand when their cache of evaluated flags must be invalidated or updated. An `on-context-set` handler can be defined which performs whatever operations are needed to reconcile the evaluated flags with the new context. The OpenFeature SDK calls this handler when the global context is modified.
```

The interface shown below has `onContextChange` as optional. I believe this is correct because not every provide will need it and it maintains backwards compatibility.

OFEP-single-context-paradigm.md
```diff
@@ -39,12 +39,14 @@ interface Provider {
   //...
 
   // a handler called by the SDK when context is modified
-  onContextChange?(oldContext: EvaluationContext, newContext: EvaluationContext): Promise<void>
+  onContextSet?(oldContext: EvaluationContext, newContext: EvaluationContext): Promise<void>
```
beeme1mr, line 42: Will `oldContext` be an empty object the first time this is called?

OFEP-single-context-paradigm.md
```diff
@@ -82,3 +84,22 @@ interface Client {
 #### Remove context mutator
 
 The global context has a one-to-one correspondence to the providers cache of evaluated flags. It's unreasonable or impossible to reconcile multiple client-level contexts with this state. The context mutator of the client must be removed.
+
+### Hook changes
+
+#### Remove return value from before stage
+
+Context cannot be modified at evaluation, so `before` stage
```
beeme1mr, line 92: This change would only apply when in "single-context" mode, right?

OFEP-single-context-paradigm.md
Kavindu-Dodan: IMO it's best not to couple initialization to context changes.

And I think we will need to expose a "single context mode", which providers can opt in and implement if they support that.

In this mode, it is best to expose an "initializer" , which performs the first initialization, caching and other provider specifics. 

This way, we can isolate implementation specifics easily.

Kavindu-Dodan: Pseudocode could look like this,

```
# initialization 
OpenFeatureAPI.setSingleContextProvider(MySingleContextProvider, Context)

# Inside setSingleContextProvider implementation
MySingleContextProvider.initialize(context)

# Inside MyProvider.initialize  implementation
// Bulk fetch of feature flags with provided context data
// Cache evaluation for offline/non-IO bound flag providing
```

OFEP-single-context-paradigm.md
toddbaert: I wouldn't describe single context as a "mode", since it can't be changed, but yes, this would only apply to SDKs which are designed to be used in single user applications, such as a web SDK.

OFEP-single-context-paradigm.md
toddbaert: Agreed, @Kavindu-Dodan . I added an initialization function in my [draft spec changes](https://github.com/open-feature/spec/pull/171/files#diff-e688c13af607ec6fef905d28a21cab877987122c01d34af42b874bb98e49b672R159).

moredip (approved):
I like the proposal overall. I particularly like that we're quite aggressively removing API surface area wrt. different ways of setting evaluation context.

I suspect at some point down the line we will look at how to support multiple independent evaluation contexts (e.g. for different teams, or for different screens in the UI). I don't *think* that this proposal is going to make it harder to add that in in the future.

OFEP-single-context-paradigm.md
```diff
@@ -0,0 +1,105 @@
...
+- flag evaluations take place against this cache, without a need to provide context (context was already used to evaluate flags in bulk)
+- Functions/methods are exposed on the SDK that signal the cache is no longer valid, and must be reconciled based on a context change. This frequently involves a network request or I/O operation.
+
+This paradigm doesn't fit well with the existing SDK, which, like most server-side SDKs, emphasizes realtime evaluation and context-per-evaluation.
+
+Though not all client side SDKs function in this way, those that do allow context per evaluation can conform to this model fairly easily.
+
+## Design
+
+To better support the "single-context" paradigm we need to define some additional handlers and add some flexibility to current APIs.
```
skyerus, line 24: How do these API changes exist alongside the current APIs? Does an author have to choose which to follow or is it possible to incorporate both designs?

toddbaert: I think it would be possible (though perhaps not recommended) to follow both. I think in most cases, client-focused SDKs would implement the static-context paradigm, while server/multi-user applications would do dynamic context. We could add guidance in the docs about this.

beeme1mr (approved):
Excited to bring OpenFeature to the client.

OFEP-single-context-paradigm.md
```diff
@@ -0,0 +1,105 @@
...
+  //...
+
+  // a handler called by the SDK when context is modified
+  onContextSet?(oldContext: EvaluationContext, newContext: EvaluationContext): Promise<void>
+
+  //...
+}
+```
+
+While the `on-context-set` handler is executing, the cache of resolved flags may be considered "stale". `Provider authors` and `application authors` should understand the consequences of evaluating flags in this state.
```
kinyoklion, line 48: I am trying to think through the process of changing the context from an application developer perspective. This method itself seems like it would generally be async and its completion would indicate that the context has chaned and flag values are now updated (Or that it has failed with some timeout). 

That doesn't go all the way to the application developer though, because this would be changing at the API (or maybe client level), which currently would not have been async.

The alternative would be that the loop would be closed via events, and maybe I need to re-read that spec, but it didn't seem concrete enough to depend on that cycle.

toddbaert: > I am trying to think through the process of changing the context from an application developer perspective. This method itself seems like it would generally be async and its completion would indicate that the context has chaned and flag values are now updated (Or that it has failed with some timeout).

Correct

> That doesn't go all the way to the application developer though, because this would be changing at the API (or maybe client level), which currently would not have been async.
> 
> The alternative would be that the loop would be closed via events, and maybe I need to re-read that spec, but it didn't seem concrete enough to depend on that cycle.

In my (very basic) PoC for the web, the `setContext` method on the global API object is a promise. This is a divergence from the server SDKs. That promise is resolved with the resolution of the `onContextSet` handler. Practically, an author would await this `setContext`.

I think I should mention something like this is the draft spec changes though: https://github.com/open-feature/spec/pull/171/files#diff-88f580a0841fa6e3905a16e608eb284f0d3441899de2f80c75a8dc7d43c13b8dR94-R98 

The alternatives that I briefly considered is to provide some explicit method for awaiting the `onContextSet`, or, as you mention, using an event, but IMO adding this behavior to the `setContext` feels more consistent with our existing paradigms, especially considering that some providers won't have to do any async operation when the context is changed, and will simply immediately resolve a promise. Interesting in your take on this, though. Do you think events are a better way to go?

kinyoklion: I personally prefer a promise returned from `setContext`. It wasn't mentioned here, so I wasn't sure if that was being considered. 

Though I wonder if we should have anything to say about timeout behavior of that promise.

toddbaert: > I personally prefer a promise returned from `setContext`. It wasn't mentioned here, so I wasn't sure if that was being considered.
> 
> Though I wonder if we should have anything to say about timeout behavior of that promise.

I could understand why a timeout would could be useful, but I'm wondering if it's something we'd want to handle in the SDK, rather than allowing the calling code to handle their own timeout with `Promise.race` or similar.

kinyoklion: Yeah, `Promise.race` is what we often suggest as a timeout to people (for SDK initialization). But it comes up extremely often, which is why I thought it may be worth mentioning. Also, when people get it wrong, their apps may never load or otherwise stall in specific conditions. May also matter if we care about resolution order to many in-flight promises to change the context.

Regardless, I think it is likely fine either way.

OFEP-single-context-paradigm.md
```diff
@@ -0,0 +1,105 @@
...
+
+#### Remove evaluation-context parameter on resolvers
+
+With all evaluation context represented in the global context, resolvers no longer need it as a parameter. Providers that require context at evaluation can simply access the global evaluation context for use in their evaluation.
```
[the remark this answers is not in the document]
toddbaert: @fabriziodemaria yes, excellent point.

I think this question also brings up whether or not we want to maintain hooks the same way in the static context paradigm. _I think the answer is yes, but I'm open to other proposals._

In terms of your specific point, I think in languages where this is a concern, it would be necessary to do some locking to prevent unpredictable behavior. We have similar locks for similar programs in some server SDKs already.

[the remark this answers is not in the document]
fabriziodemaria: One proposal could be for the Provider's resolve methods to still accept the `ctx` from the SDK Client, rather than having the Provider internally read the global property; this way we can at least ensure that the same `ctx` is used in any one evaluation and its corresponding hooks. As a side note, I also think that keeping the `ctx` input argument in the Provider interface could streamline slightly writing tests for the Providers, at least in certain programming languages; on top of that, this would make the Provider's implementation perhaps slightly more agnostic of the SDK's global properties, which sounds a little cleaner to work with. What do you think?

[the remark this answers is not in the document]
toddbaert: @fabriziodemaria I think that is a good idea because it _probably_ reduces the need for locking and gives us more control, and it also is more consistent, as you point out. I will make a change here and also on the draft [PR spec](https://github.com/open-feature/spec/pull/171).

OFEP-single-context-paradigm.md
toddbaert: I've done this: https://github.com/open-feature/ofep/pull/41/commits/4083c83bd37ae4f37dc02155b27b256a6ac937bc

OFEP-single-context-paradigm.md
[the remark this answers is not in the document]
toddbaert: I've made this change here: https://github.com/open-feature/ofep/pull/41/commits/4083c83bd37ae4f37dc02155b27b256a6ac937bc

OFEP-single-context-paradigm.md
fabriziodemaria: Following up from @Kavindu-Dodan suggestion:

Should we _only_ expose a `setProvider` API that requires to pass in both the new provider and a new non-null context? Can we foresee any use-case where users would want to set a new provider without any global context available?

The reason why I am asking: as long as it's possible to set a new provider with global context being null, the `initialize` and `onContextSet` functions would need to pass the current context _optionally_ to the provider (for the languages that support optional types).

The options being (pseudocode):
- Allow `setProvider(provider)` and pass the context optionally in followup functions, e.g. `initialize(context?)`
- Only allow `setProvider(provider, context)` and pass the context strictly not-null in followup functions, e.g. `initialize(context)`

Kavindu-Dodan: > Can we foresee any use-case where users would want to set a new provider without any global context available?

I think this is a valid use case. For example, at the application startup, the provider can be set globally. Yet the context can be empty/null at this stage as context details are unavailable. But once the context is ready/known (ex:- user login, region detection, page visit...), the context can be initialized. 

Contrary to my pseudocode, it makes sense to go with `initialize(context)` followed by `setProvider(provider)`.

s-sen:
An initialization API was added to the java sdk provider interface which does not have any context parameters. Is there any discussion that describes why that was added and why there is no context parameter?

PR #44, opened 2023-01-10
Ofep/flag metadata to applications 

Kavindu-Dodan:
## This PR
Proposes to introduce flag metadata exposure to the application layer.

With flag metadata such as key, flag type, application developers get the ability to build various validations, and tools on-top of already existing APIs.


OFEP-flag-metadata-to-applications.md
```diff
@@ -0,0 +1,80 @@
...
+
+- flag key 
+- flag type 
+
+Other vendor specific, non-sensitive data can also be present and they can be included in a dedicated `data` segment.
+
+From a JSON representation, this payload could look like below,
+
+```json
+{
```
skyerus, line 31: Is this suggesting that the metadata would be stored in the flag configuration?

OFEP-flag-metadata-to-applications.md
```diff
@@ -0,0 +1,80 @@
...
+For example, cross-validations of flags used by an application using unit tests without the need for evaluation API 
+or a connection to the flag management system.
+
+## Proposal
+
+The content of metadata payload can vary from provider to provider. However, from OpenFeature specification perspective,
+it should be possible to enforce the following, 
+
+- flag key 
+- flag type
```
skyerus, line 24: Implying that flag value should not be returned within the metadata? It may be prudent to explicitly state this

OFEP-flag-metadata-to-applications.md
```diff
@@ -0,0 +1,80 @@
...
+
+
+## Background
+
+At the time of writing this document, OpenFeature API supports metadata for providers and client. With a similar 
+approach, it is desirable to expose flag metadata to the application layer.
+
+This metadata then can be used by application authors for various purposes such as validations, tooling and testing.
+For example, cross-validations of flags used by an application using unit tests without the need for evaluation API 
+or a connection to the flag management system.
```
skyerus, line 16: How would a client retrieve the metadata if not from the flag management system?

skyerus, line 16:
```suggestion
or a direct connection to the flag management system.
```

OFEP-flag-metadata-to-applications.md
Kavindu-Dodan: No, I used JSON here merely as a representation. Vendors can use any storage, but API will at least require "key" and "type" in the result.

beeme1mr:
I think this proposal is a bit out of scope given the current focus of OpenFeature. I can see how this could be valuable but primarily for testing and CI processes. In my opinion, this shouldn't be included in the SDKs, at least not at this time.

moredip:
I agree with @beeme1mr that this feels out of scope *for now*, but @Kavindu-Dodan I would love to understand the use case more. 

What would you imagine an OF consumer (an [Application Author](https://docs.openfeature.dev/docs/specification/glossary/#application-author) or an [Application Integrator](https://docs.openfeature.dev/docs/specification/glossary/#application-integrator), or some other persona) using this capability for, if we built it?

Kavindu-Dodan:
Thanks for the feedback @beeme1mr  & @moredip . Yes I want to think more about "how" different entities can benefit from flag metadata. I will think more about this and come back if I see better use cases than merely using it for testing and validations.

Issue #49, opened 2023-02-24
Global providers and context


toddbaert:
Particularly with out upcoming support for client-side, I'm concerned about the global state (`provider` and global `context`) in our SDKs. We may want to provide some alternative APIs that don't rely on these global functions/state. They provide benefits, but also come with costs and risks.

Benefits:

- easy, central configuration
- configuration can be leveraged by 3rd party libraries targeting the evaluation API (a provider set globally will be used by evaluations in a 3rd party library without requiring any explicit openfeature configuration in that library)

Costs/Risks:

- testing global singletons is tricky and error prone
- unexpected bugs, side effects, reentrancy issues (global state can be problematic)
- lack of flexibility (two teams contributing components to the same webapp would not be able to use different providers)

I'd like to explore solutions to this. One I've been considering is to simply add a `setProvider` method on the OpenFeature client. This would allow configuration of a provider simply for this client, not impacting the global API or other clients.

This is fairly straightforward in the server. In the case of the client, more might need to be considered. In our working client proposal, we don't support passing a context at evaluation time. Instead, the context is maintained on the global API object and changes to it have side-effects in providers (see [here](https://github.com/open-feature/ofep/pull/41) for details). If we'd like to provide alternatives to global state, we'd also need to allow for setting context on the client in client-side SDKs. To clearly scope the impact of changing the context in this case, I propose that we only allow the client context to be set immutably along with the client provider. The API might look like:

```java
client.setProvider(provider, context); // pass optional provider and context
```

or perhaps when generating the client: 

```java
Client client = OpenFeature.getClient(provider, context); // pass optional provider and context
```
This is just one idea... and perhaps my concerns are unfounded. Let me know your thoughts.

justinabrahms:
I support adding client-side state and providers. I don't support removing global state which cascades down into the client.

As an example for a good use of global state, I want to include things like "Which environment is this deployed in?" or "Which server farm is this in?" as context in my calls.

toddbaert:
> I support adding client-side state and providers. I don't support removing global state which cascades down into the client.

>As an example for a good use of global state, I want to include things like "Which environment is this deployed in?" or "Which server farm is this in?" as context in my calls.

OK. Ya, I think we should definitely _keep_ the global stuff. I'm only advocating adding the others. I think this is a good point though, I will add this to my initial comment.

moredip:
[...]

PR #56, opened 2023-03-10
Provider to client mappings

justinabrahms:
refs: #49


OFEP-provider-client-mapping
```diff
@@ -0,0 +1,33 @@
+## Provider to client mapping
+
+## State: APPROVED
+
+The goal of this OFEP is to describe a way that application authors can selectively determine which provider a given client uses. This means that they should be able to set a provider for a library that is different than what they use in their main app.
```
toddbaert, line 5: @thiyagu06 this might apply to your situation, though I think in your case it's more about different parts of a large application.

OFEP-provider-client-mapping
```diff
@@ -0,0 +1,33 @@
...
+
+## Proposal
+
+To address that, we will allow application authors to provide mappings of client names to a provider of their choosing. This should allow a great degree of flexibility for app authors to choose the flag-config store that works for that particular use-case.
+
+### Example Implementation
+
+Clients may optionally use [a named client](https://github.com/open-feature/spec/blob/74c373e089ad77bf8cac84f3d93c00c945ff3a8a/specification/sections/01-flag-evaluation.md?plain=1#L60), which we already support. If they don't pick a name, it falls through to a "root" provider.
```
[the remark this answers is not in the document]
toddbaert: I _think_ this is slightly more accurate, if you're talking about the "getClient" call.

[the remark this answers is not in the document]
toddbaert: I'm not sure if it's also worth mentioning that not specifying a named client will always generate a `new` client?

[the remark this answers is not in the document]
justinabrahms:
```suggestion
Authors may optionally use [a named client](https://github.com/open-feature/spec/blob/74c373e089ad77bf8cac84f3d93c00c945ff3a8a/specification/sections/01-flag-evaluation.md?plain=1#L60), which we already support. If they don't pick a name, it falls through to the configured "root" provider (or no provider if none are set).
```

OFEP-provider-client-mapping
```diff
@@ -0,0 +1,33 @@
...
+# example library code
+var client = OpenFeature.getInstance().getClient('my-sweet-library')
+if client.getBooleanValue('redesign-enabled?', false) {
+  # ...
+}
+```
+
+```
+# application author
+OpenFeature.setProvider(OpenFeature.ROOT_PROVIDER, new FlagDProvider())
```
thiyagu06, line 29: should we call it as `ROOT_PROVIDER`? when we say `ROOT_PROVIDER`, I'm thinking that my flag evaluation will start from root if didn't specify a client name and goes down to evaluate the flag with all other providers. (root/leaf or parent/child relationship). I would prefer calling it as `DEFAULT_PROVIDER`  to avoid the confusion?

OFEP-provider-client-mapping
```diff
@@ -0,0 +1,33 @@
...
+
+To address that, we will allow application authors to provide mappings of client names to a provider of their choosing. This should allow a great degree of flexibility for app authors to choose the flag-config store that works for that particular use-case.
+
+### Example Implementation
+
+Authors may optionally use [a named client](https://github.com/open-feature/spec/blob/74c373e089ad77bf8cac84f3d93c00c945ff3a8a/specification/sections/01-flag-evaluation.md?plain=1#L60), which we already support. If they don't pick a name, it falls through to the configured "default" provider (or no provider if none are set).
+
+```java
+# example library code
+var client = OpenFeature.getInstance().getClient('my-sweet-library')
```
thiyagu06, line 21: Do we need to document about the SDK behaviour when no provider exists for the given client name?

justinabrahms: We will in the spec, but as mentioned.. it falls through to the default provider.

To be clear, 'my sweet library' is the name of the client, not the provider.

thiyagu06: Just to make sure that, I understand it properly, If I do `OpenFeature.getInstance().getClient('not_exists')`, you meant to say that it fallback to the default provider.  If so, it's better we let the application author to decide on the behaviour they want if the expected client is not available rather than we fall back to default provider. 

```
OpenFeature.getInstance().getClient() or OpenFeature.getInstance().getClient(OpenFeature.DEFAULT_PROVIDER) -> FlagDProvider
OpenFeature.getInstance().getClient('my-sweet-library') -> FileBasedProvider 
OpenFeature.getInstance().getClient('not_exists') -> null
``

**Review by thomaspoignant — CHANGES_REQUESTED** [review]
Quick feedback on the format, you're file is not finishing by `.md` so it does not format well in Github.

File: OFEP-provider-client-mapping (hunk 4 above)
Comment (justinabrahms) [reply OFEP-provider-client-mapping:21]: You're not quite getting what I'm saying. The `getClient()` function does not take a provider name. It takes an optional client name.  So to reuse your example:

```
OpenFeature.getInstance().getClient() -> flagD provider
OpenFeature.getInstance().getClient(OpenFeature.DEFAULT_PROVIDER) // this is nonsensical. It doesn't reference a provider
OpenFeature.getInstance().getClient('my-sweet-library') -> FileBasedProvider 
OpenFeature.getInstance().getClient('not_exists') -> flagD provider
```

Comment (thiyagu06) [reply OFEP-provider-client-mapping]: > OpenFeature.getInstance().getClient(OpenFeature.DEFAULT_PROVIDER) // this is nonsensical. It doesn't reference a provider

If I could set a provider with this (as per the example) OpenFeature.setProvider(OpenFeature.DEFAULT_PROVIDER, new FlagDProvider()) where OpenFeature.DEFAULT_PROVIDER is String constant, then why it is nonsensical get the client with the same constant?

> OpenFeature.getInstance().getClient('my-sweet-library') -> FileBasedProvider 

I understand that we're mapping the client name with the provider name which is used here `OpenFeature.setProvider('my-sweet-library', new FileBasedProvider('path/to/myfile'))` to get the client with specific feature config store

Is there a way I could get `FileBasedProvider` as feature config store with `OpenFeature.getInstance().getClient('not_exists')`. If not, then my client name should be always `my-sweet-library` to get `FileBasedProvider` as my feature config store. 

Hope I'm clear to you.

Comment (justinabrahms) [reply OFEP-provider-client-mapping]: > > OpenFeature.getInstance().getClient(OpenFeature.DEFAULT_PROVIDER) // this is nonsensical. It doesn't reference a provider
>
> If I could set a provider with this (as per the example) OpenFeature.setProvider(OpenFeature.DEFAULT_PROVIDER, new FlagDProvider()) where OpenFeature.DEFAULT_PROVIDER is String constant, then why it is nonsensical to get the client with the same constant?

It is syntactically valid to pass in the DEFAULT_PROVIDER constant, however it's confusing the nature of things. What you pass to a client is the name of the client, not an indication of which providers to use.

---



> >  OpenFeature.getInstance().getClient('not_exists') -> flagD provider
>
> Is there a way I could get FileBasedProvider as feature config store with OpenFeature.getInstance().getClient('not_exists'). If not, then my client name should be always my-sweet-library to get FileBasedProvider as my feature config store. Is that correct?

This is similarly confusing things. `not_exists` returns the configured default provider. This isn't because `not_exists` is some special string, but because we didn't configure a mapping of client name to a different provider.

While it really makes things confusing for this example, an application author could configure `setProvider('not_exists', myFileProvider)` and it would return the file provider.  The client name to choose is the decision of people using clients (library authors and application authors). How to actually map those names to feature providers is solely the decision of the application author when they configure things.

Comment (toddbaert) [reply OFEP-provider-client-mapping]: @justinabrahms I think I like this proposal, but I had similar confusion to @thiyagu06 initially trying to grok this proposal, though that could be because my proposal was to create a provider map.

I wonder if there's any tiny changes we can make to avoid this confusion. I also wonder if we would want to support identifying more than one client name to associate with the provider in a single call.

**Comment by beeme1mr** [comment]
@justinabrahms how would you feel about something like this?

```typescript
import { OpenFeature } from '@openfeature/js-sdk';

OpenFeature.setProvider(new MyGlobalProvider());

const client = OpenFeature.getClient();
client.overrideProvider(new MyClientProvider());
```

The idea would be to only allow a single global provider at a time but allow clients to optionally set a new provider that only affects themselves.

justinabrahms:
@beeme1mr A different way to say that:

1. libraries must never create a client. They must be passed a client.
2. If you need to juggle multiple providers (e.g. flagsmith for some stuff, flagd for others, file-based for a third), you will need to juggle multiple clients.

I don't love either of those.

OFEP-provider-client-mapping
thiyagu06: @toddbaert , Yes, the proposal would address my issue. Still I have concerns which may be related to implementation of SDK.

1. when I call `setProvider('my-sweet-library', myFileProvider)`, it seems that we are mixing up the `client_name` with provider instances. This is the reason for all my confusion. I'm not quite sure how easy to understand this behaviour for the wider audience as well.

I'm also wondering can we reuse the  provider name from the metadata https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/FeatureProvider.java#L10. so when registering the provider we can resolve the name from metadata object to create a map of providers and then provide an API like below to get the client for specific provider `OpenFeature.getInstance().getClientForProvider('flagd')` ( we can come up with better name 🤔 )

You can refer this to get some inspiration on how open telemetry is solving similar issue. https://github.com/open-telemetry/oteps/blob/main/text/0016-named-tracers.md#explanation

2. If I have only one provider in my application, how does the SDK will looks like? Are we going to force the developer to set the default provider like this `OpenFeature.setProvider(OpenFeature.DEFAULT_PROVIDER, new FlagDProvider())`  or  if the name is not provided while registering the provider  like this `OpenFeature.setProvider(new FlagDProvider())`, then the `FlagDProvider` will become a default provider?

tcarrio (approved):
All good here. Brings utility to the existing named clients (but also will be a breaking change, just to note).

OFEP-provider-client-mapping
toddbaert: > 1. when I call `setProvider('my-sweet-library', myFileProvider)`, it seems that we are mixing up the `client_name` with provider instances. This is the reason for all my confusion. I'm not quite sure how easy to understand this behaviour for the wider audience as well.

I share this concern because I had the same exact confusion earlier. On the other hand, I think this is better than forcing application authors to know about provider mappings. I think that is a concern that @moredip and @justinabrahms both have, and I agree with it.

Part of it could be naming? Improved naming can help communicate concepts a lot. Perhaps instead of client name, we want to find a better name... like `scope`? cc @justinabrahms @moredip 

> 2. If I have only one provider in my application, how does the SDK will looks like? Are we going to force the developer to set the default provider like this `OpenFeature.setProvider(OpenFeature.DEFAULT_PROVIDER, new FlagDProvider())`  or  if the name is not provided while registering the provider  like this `OpenFeature.setProvider(new FlagDProvider())`, then the `FlagDProvider` will become a default provider?

I think if you don't mention a client name (maybe `scope` moving forward) you the registered provider stays global and the current behavior persists.

thiyagu06: Thanks @toddbaert. This make sense to me now 👍

justinabrahms:
> All good here. Brings utility to the existing named clients (but also will be a breaking change, just to note).

My intent would be to still allow `setProvider(provider)` to set the default one, which should keep it backwards compatible, yes?

justinabrahms:
Here's a prototype: https://github.com/open-feature/java-sdk/pull/388

toddbaert:
> Here's a prototype: [open-feature/java-sdk#388](https://github.com/open-feature/java-sdk/pull/388)

Thanks @justinabrahms I'll take a look at this tomorrow!

toddbaert:
> Here's a prototype: [open-feature/java-sdk#388](https://github.com/open-feature/java-sdk/pull/388)

@justinabrahms This makes sense to me. I have a couple thoughts/questions:

- How do you feel about using "scope" to refer to the "name" for a client. So `getProviderForClientOrDefault(String name)` would change to `getProviderForScopeOrDefault(String scope)`? I know it's just naming, but I think referring to this as a "scope" adds clarify, and even allows us to future proof things a bit; there may be other things we want to "scope" in this way. See earlier discussion [here](https://github.com/open-feature/ofep/pull/56#discussion_r1142280472). We could add this term to our documentation and gloassary.
- I suppose initialization and events (both not yet completely spec'd out) would be scoped similarly? So change events emitted in a particular provider only impact clients in that provider scope? and setting a new provider for a given scope would shut down the previous one?
- Would API shutdown (discussed [here](https://github.com/open-feature/spec/pull/179/files#diff-1443054d965453435ba3da2907dabbc282d2f808c3a64800f10e05fab8351b1aR223-R229)) still be global? Would there be a need to shutdown only certain providers? I'm leaning towards no, since I think the shutdown is really only useful for terminating the application, in which case all providers should probably be stopped.

cc @thiyagu06 @tcarrio @beeme1mr @thomaspoignant

justinabrahms:
@toddbaert I dislike using the term 'scope' for the client name. It introduces another term, which I think confuses things. What we're really talking about is binding a client to a specific provider. It's not some "scoping key" that is being bound. It's the named client.

I would expect that initialization would happen for any provider that's registered. I'm unclear about how events would be supported. One option would be a single event stream with a parameter of what is effected or we could have multiple event streams. I don't feel strongly.

I think API shutdown would be global.

thomaspoignant:
I can agree with @justinabrahms I don't really like the word scope in this context it can be misleading and it is not completely obvious that we are talking about the client name.

toddbaert:
> I would expect that initialization would happen for any provider that's registered. I'm unclear about how events would be supported. One option would be a single event stream with a parameter of what is effected or we could have multiple event streams. I don't feel strongly.

@justinabrahms my main concern was to confirm that if provider associated with client X gets a "flags change" event, only client X would get that event, right?

justinabrahms:
@toddbaert I'm not fully up to speed on the events stuff. I don't think there's a technical reason why the thing you're saying can't work though.

toddbaert (approved):

thomaspoignant (approved):

PR #59, opened 2023-03-20
feat: add proposal for ofo flagd service

skyerus:
## This PR

OFEP for ofo flagd service

### Related Issues

https://github.com/open-feature/open-feature-operator/issues/371


beeme1mr (approved):
Simple and to the point. This sounds like a nice, flexible deployment option.

OFEP-ofo-flagd-service.md
```diff
@@ -0,0 +1,14 @@
...
+Currently, OpenFeature Operator (OFO) manages the deployment of flagd by appending it to a pod's containers (sidecar pattern), thereby allowing containers within the pod to route to it. This OFEP campaigns for the extension of OFO to manage the deployment of externally accessible flagd (e.g. from client side applications, or even server side applications that don't want to use the sidecar pattern).
+
+## Background
+
+The driving force behind this is to simplify the deployment of flagd for use by client side applications. An [OFEP was drafted](./OFEP-ofo-flagd-client-support.md) to achieve this but subsequently withdrawn (reasons noted within). This OFEP recognises the limitations of the previous, presenting a modular solution with a broader scope.
+
+## Proposal
+
+Introduce a FlagdService custom resource definition (CRD) and controller.
+The controller uses the configuration defined within the custom resource (CR) to create a Service and a Deployment of flagd (backed by the Service). This is a common deployment pattern permitting access by any component that routes to the created Service (e.g. Ingress/Load Balancer).
```
toddbaert, line 14: You might want to very briefly contrast this with the "injected sidecar" model (flagd+fronting service vs flagd injected into workload pod).

toddbaert (approved):
Approved with small suggestion.

OFEP-ofo-flagd-service.md
thisthat: I like the suggestion of @toddbaert
I would suggest highlighting as part of this OFEP the new set of RBAC required to introduce this.  
@skyerus do you plan to allow the CR to specify also the ns in which the deployment & service are created?

thisthat (approved):
As a Platform Eng, I would have some reservation deploying an operator that can control K8s Services in any namespace. In the implementation details, I would suggest to enable/disable this behavior via (Open)Feauture Flag 😉 In the docs, we could explain that if this is too much permissions, this behavior can also be done manually by applying a manifest with the svc

OFEP-ofo-flagd-service.md
```diff
@@ -0,0 +1,33 @@
...
+
+Currently, OpenFeature Operator (OFO) manages the deployment of flagd by appending it to a pod's containers (sidecar pattern), thereby allowing containers within the pod to route to it. This OFEP campaigns for the extension of OFO to manage the deployment of externally accessible flagd (e.g. from client side applications, or even server side applications that don't want to use the sidecar pattern).
+
+## Background
+
+The driving force behind this is to simplify the deployment of flagd for use by client side applications. An [OFEP was drafted](./OFEP-ofo-flagd-client-support.md) to achieve this but subsequently withdrawn (reasons noted within). This OFEP recognises the limitations of the previous, presenting a modular solution with a broader scope.
+
+## Proposal
+
+Introduce a FlagdService custom resource definition (CRD) and controller.
```
beeme1mr, line 13: I wonder if we could make this a bit more generic. Most flag vendors offer a remote evaluation service of some kind. We could start by only supporting flagd but it may be nice to have the option of adding additional tools if there's community interest.

OFEP-ofo-flag-service.md
```diff
@@ -0,0 +1,42 @@
+## flag service deployment driven by OpenFeature Operator
+
+## State: DRAFTING
+
+Currently, OpenFeature Operator (OFO) manages the deployment of a flag provider (e.g. flagd) by appending it to a pod's containers (sidecar pattern), thereby allowing containers within the pod to route to it. This OFEP campaigns for the extension of OFO to manage the deployment of an externally accessible flag provider (e.g. from client side applications, or even server side applications that don't want to use the sidecar pattern).
```
AlexsJones, line 5: Can we discuss how this would work in principal? Would it be limited to the same Kubernetes cluster or network segment? The first thought is around connectivity and the challenges there. Would say its the users responsibility to provide a flat network?

skyerus: Limiting the resources created to a Deployment & Service means that the networking beyond that is left as a responsibility to the users.
The deployment could be created in whichever namespace is provided but the Service CRUD is intentionally limited to OFO's namespace to minimise security risks.

In my mind, once the workload has been deployed (flagd) the Kubernetes primitives (i.e. Deployments) take charge of the lifecycle of the flag provider.

beeme1mr: Perhaps you could rephrase `This OFEP campaigns for the extension of OFO to manage the deployment of an externally accessible flag provider`  to something like `This OFEP campaigns for the extension of OFO to manage flag providers by abstracting a Kubernetes Deployment resource`.

skyerus: Great suggestion, done

skyerus:
Further POC work has led me to discover that services cannot expose deployments across namespaces.

The simplest circumvention to this is to enforce **both** the Deployment and Service to be created in OFO's namespace.
If this isn't extensive enough, an alternative could be to provide a list of allowed namespaces to OFO in which Service CRUD is permitted.
I prefer the former due to its simplicity. If OFO's namespace ends up being too limiting it can be built upon in the future.

Thoughts @toddbaert @beeme1mr @thisthat @AlexsJones ?

AlexsJones:
I am really confused, I left a comment on this and it's been merged but I can't find the comment

beeme1mr:
> I am really confused, I left a comment on this and it's been merged but I can't find the comment

Here's a link to your comment: https://github.com/open-feature/ofep/pull/59#discussion_r1145018209

skyerus:
> I am really confused, I left a comment on this and it's been merged but I can't find the comment

Your comment is still visible to me: https://github.com/open-feature/ofep/pull/59#discussion_r1145018209

> Okay I'll add what I think I wrote the other day:
> 
> > The driving force behind this is to simplify the deployment of flag providers for use by client side applications
> 
> I think this is somewhat misleading, this doesn't make deploying clientside application deployment simpler, it makes it prescriptive, it's providing a way we think is best to do it, rather than letting people do it themself.

People are still able to do it themselves if they don't fit to this prescription but imo the Service behind Deployment is common enough to warrant this enhancement.

> 
> > In contrast, the described FlagService pattern permits an externally routable flag provider.
> 
> Why not simply route from the container that has the client SDK inside itself?

Can you elaborate further on this please? Not sure I fully understand.

toddbaert:
@AlexsJones I think you may be misunderstand what we mean when we say "client flags".

> Why not simply route from the container that has the client SDK inside itself?

In client side flagging, _there is no container with an SDK_. The SDK and the provider exist in _client code_ (an iOS app, web browser, etc).

toddbaert:
See my previous comment [here](https://github.com/open-feature/open-feature-operator/issues/371#issuecomment-1472457310) for more background on the paradigm.

PR #62, opened 2023-05-19
feat: ofep for metric hooks

Kavindu-Dodan:
## This PR

Introduce an enhancement proposal to introduce OpenTelemetry metric hooks. 

Consider checking PR [1] for a POC implementation in Go sdk. 

Besides, see below Prometheus samples for metrics,

**feature_flag.evaluation_requests_total**
image

**feature_flag.evaluation_success_total**
image

**feature_flag.evaluation_error_total**
image

**feature_flag.evaluation_requests_total**
image

[1] - https://github.com/open-feature/go-sdk-contrib/pull/217

### DIFF
--- a/OFEP-metric-hooks.md
+++ b/OFEP-metric-hooks.md
@@ -1,0 +1,67 @@
+## OFEP-single-context-paradigm
+
+## State: PENDING REVIEW
+
+This OFEP propose to introduce OpenFeature hook for OpenTelemetry metrics.
+
+## Background
+
+We already have OpenTelemetry Span support through hooks. Similarly, we can provide a dedicated hook for 
+OpenTelemetry 
+metrics. Providing telemetry data out of the box will make OpenFeature attractive for both users and vendors. 
+
+## Proposal
+
+The proposal here is to define a set of metrics that can be used with different hook stages. For example, error metrics 
+can e collected at hook's `error` stage. Going with this background, I propose the following metrics.
+
+### evaluationRequests
+
+A counter[1] based metric to calculate the number of flag evaluation requests. This is recorded at `before` stage of 
+the hook.
+
+### evaluationSuccess
+
+A counter[1] based metric to calculate the number of successful flag evaluations. This is recorded at `after` stage of 
+the hook.
+
+### evaluationError
+
+A counter[1] based metric to calculate the number of failed flag evaluations. This is recorded at `error` stage of
+the hook.
+
+### evaluationActive
+
+An UpDownCounter[2] based metric to calculate the number of active flag evaluations currently going through 
+OpenFeature SDK. This is increased at `before` stage and decreased at `finally` stage. Given the evaluation is fast,
+the value observed here can be 0. However, when there are bottlenecks or provider slowdowns, this will be a 
+non-zero value. 
+
+### Semantic conventions 
+
+All above metrics should carry OpenTelemetry semantic convention defined attributes [3] so that any metric processor 
+can reliably process metrics produced by the hook.
+
+## Example
+
+Consider following coding example in Go. 
+
+```go
+
+    // Reader should be derived or injected 
+	var reader metric.Reader
+
+	// Derive metric hook from reader
+	metricsHook, := hooks.NewMetricsHook(reader)
+
+	// Register OpenFeature API hooks
+	openfeature.AddHooks(metricsHook)
+```
+
+### References
+
+[1] - https://opentelemetry.io/docs/specs/otel/metrics/api/#counter
+
+[2] - https://opentelemetry.io/docs/specs/otel/metrics/api/#updowncounter
+
+[3] - https://opentelemetry.io/docs/specs/otel/logs/semantic_conventions/feature-flags/ 
--- a/OFEP-metric-hooks.md
+++ b/OFEP-metric-hooks.md
@@ -6,12 +6,12 @@ This OFEP propose to introduce OpenFeature hook for OpenTelemetry metrics.
 
 ## Background
 
-We already have OpenTelemetry Span support through hooks. Similarly, we can provide a dedicated hook for OpenTelemetry 
-metrics. Providing telemetry data out of the box will make OpenFeature attractive for both users and vendors. 
+We already have OpenTelemetry Span support through hooks. Similarly, we can provide a dedicated hook for OpenTelemetry
+metrics. Providing telemetry data out of the box will make OpenFeature attractive for both users and vendors.
 
 ## Proposal
 
-The proposal here is to define a set of metrics that can be used with different hook stages. For example, error metrics 
+The proposal here is to define a set of metrics that can be used with different hook stages. For example, error metrics
 can be collected at the hook's `error` stage. Going with this background, I propose the following metrics.
 
 ## Metrics
@@ -20,26 +20,26 @@ All metrics defined through this proposal carry OpenTelemetry semantic conventio
 
 - feature_flag.key: The unique identifier of the feature flag
 - feature_flag.provider_name: The name of the service provider that performs the flag evaluation
-- feature_flag.variant:	SHOULD be a semantic identifier for a value. If one is unavailable, a stringified version of 
+- feature_flag.variant:    SHOULD be a semantic identifier for a value. If one is unavailable, a stringified version of
   the value can be used
 
 Given below is the list of metrics proposed and their usage,
 
 ### feature_flag.evaluation_requests
 
-A counter[2] based metric to calculate the number of flag evaluation requests. This is recorded at `before` stage of 
+A counter[2] based metric to calculate the number of flag evaluation requests. This is recorded at `before` stage of
 the hook.
 
 This metric is useful to understand flag evaluation requests received through OpenFeature SDK.
 
 ### feature_flag.evaluation_success
 
-A counter[2] based metric to calculate the number of successful flag evaluations. This is recorded at `after` stage of 
+A counter[2] based metric to calculate the number of successful flag evaluations. This is recorded at `after` stage of
 the hook.
 
 This metric is useful to understand successful flag evaluations. This metric contain following extra dimension(s),
 
-- reason : evaluation reason extracted from flag resolution details[3] 
+- reason : evaluation reason extracted from flag resolution details[3]
 
 ### feature_flag.evaluation_error
 
@@ -52,26 +52,26 @@ This metric is useful to understand flag evaluation errors. This metric contain 
 
 ### feature_flag.evaluation_active
 
-An UpDownCounter[4] based metric to calculate the number of active flag evaluations currently going through 
+An UpDownCounter[4] based metric to calculate the number of active flag evaluations currently going through
 OpenFeature SDK. This is increased at `before` stage and decreased at `finally` stage.
 
 Given the evaluation is fast, the value observed here can be 0. However, when there are bottlenecks or provider
-slowdowns, this will be a non-zero value. 
+slowdowns, this will be a non-zero value.
 
 ## Example
 
-Consider following coding example in Go. 
+Consider following coding example in Go.
 
 ```go
 
-        // Reader should be derived or injected 
-	var reader metric.Reader
+// Reader should be derived or injected 
+var reader metric.Reader
 
-	// Derive metric hook from reader
-	metricsHook := hooks.NewMetricsHook(reader)
+// Derive metric hook from reader
+metricsHook := hooks.NewMetricsHook(reader)
 
-	// Register OpenFeature API hooks
-	openfeature.AddHooks(metricsHook)
+// Register OpenFeature API hooks
+openfeature.AddHooks(metricsHook)
 ```
 
 ### References
--- a/OFEP-metric-hooks.md
+++ b/OFEP-metric-hooks.md
@@ -37,7 +37,7 @@ This metric is useful to understand flag evaluation requests received through Op
 A counter[2] based metric to calculate the number of successful flag evaluations. This is recorded at `after` stage of
 the hook.
 
-This metric is useful to understand successful flag evaluations. This metric contain following extra dimension(s),
+This metric is useful for understanding successful flag evaluations. This metric contains the following extra dimension(s),
 
 - reason : evaluation reason extracted from flag resolution details[3]
 
--- a/OFEP-metric-hooks.md
+++ b/OFEP-metric-hooks.md
@@ -43,7 +43,7 @@ This metric is useful for understanding successful flag evaluations. This metric
 
 ### feature_flag.evaluation_error_total
 
-A counter[2] based metric to calculate the number of failed flag evaluations. This is recorded at `error` stage of
+A counter[2] based metric to calculate the number of failed flag evaluations. This is recorded at the `error` stage of
 the hook.
 
 This metric is useful to understand flag evaluation errors. This metric contain following extra dimension(s),
--- a/OFEP-metric-hooks.md
+++ b/OFEP-metric-hooks.md
@@ -1,6 +1,6 @@
 ## OFEP-hook-for-metrics
 
-## State: PENDING REVIEW
+## State: APPROVED
 
 This OFEP proposes to introduce an OpenFeature hook for OpenTelemetry metrics.
 
--- a/OFEP-metric-hooks.md
+++ b/OFEP-metric-hooks.md
@@ -2,7 +2,7 @@
 
 ## State: PENDING REVIEW
 
-This OFEP propose to introduce OpenFeature hook for OpenTelemetry metrics.
+This OFEP proposes to introduce an OpenFeature hook for OpenTelemetry metrics.
 
 ## Background
 
--- a/OFEP-metric-hooks.md
+++ b/OFEP-metric-hooks.md
@@ -46,7 +46,7 @@ This metric is useful for understanding successful flag evaluations. This metric
 A counter[2] based metric to calculate the number of failed flag evaluations. This is recorded at the `error` stage of
 the hook.
 
-This metric is useful to understand flag evaluation errors. This metric contain following extra dimension(s),
+This metric is useful for understanding flag evaluation errors. This metric contains the following extra dimension(s),
 
 - exception : error/exception message extracted from the evaluation error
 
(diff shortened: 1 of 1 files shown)


OFEP-metric-hooks.md
```diff
@@ -0,0 +1,67 @@
+## OFEP-single-context-paradigm
```
beeme1mr, line 1: This looks like a copy/paste issue

OFEP-metric-hooks.md
```diff
@@ -0,0 +1,67 @@
...
+## Background
+
+We already have OpenTelemetry Span support through hooks. Similarly, we can provide a dedicated hook for 
+OpenTelemetry 
+metrics. Providing telemetry data out of the box will make OpenFeature attractive for both users and vendors. 
+
+## Proposal
+
+The proposal here is to define a set of metrics that can be used with different hook stages. For example, error metrics 
+can e collected at hook's `error` stage. Going with this background, I propose the following metrics.
```
thisthat, line 16:
```suggestion
can be collected at the hook's `error` stage. Going with this background, I propose the following metrics.
```

OFEP-metric-hooks.md
```diff
@@ -0,0 +1,67 @@
...
+### evaluationActive
+
+An UpDownCounter[2] based metric to calculate the number of active flag evaluations currently going through 
+OpenFeature SDK. This is increased at `before` stage and decreased at `finally` stage. Given the evaluation is fast,
+the value observed here can be 0. However, when there are bottlenecks or provider slowdowns, this will be a 
+non-zero value. 
+
+### Semantic conventions 
+
+All above metrics should carry OpenTelemetry semantic convention defined attributes [3] so that any metric processor
```
thisthat, line 42:
```suggestion
All above metrics should carry OpenTelemetry semantic conventions defined attributes [3] so that any metric processor 
```

OFEP-metric-hooks.md
```diff
@@ -0,0 +1,67 @@
...
+
+Consider following coding example in Go. 
+
+```go
+
+    // Reader should be derived or injected 
+	var reader metric.Reader
+
+	// Derive metric hook from reader
+	metricsHook, := hooks.NewMetricsHook(reader)
```
thisthat, line 55: or do you expect also an error to be returned? 
```suggestion
	metricsHook := hooks.NewMetricsHook(reader)
```

OFEP-metric-hooks.md
```diff
@@ -0,0 +1,67 @@
...
+can reliably process metrics produced by the hook.
+
+## Example
+
+Consider following coding example in Go. 
+
+```go
+
+    // Reader should be derived or injected 
+	var reader metric.Reader
```
thisthat, line 52:
```suggestion
        // Reader should be derived or injected 
	var reader metric.Reader
```

thisthat (commented):
Should we define the naming of the exposed metrics as part of this OFEP or would be this covered in the following up spec change?

OFEP-metric-hooks.md
Kavindu-Dodan: Thank you, yes, this was while extracting the format of the ofep.

Kavindu-Dodan: Resolved with the latest commit

Kavindu-Dodan:
> Should we define the naming of the exposed metrics as part of this OFEP or would be this covered in the following up spec change?

Right now there's no OpenFeature spec section defining semantic conventions. So if we see the need, we will need to introduce this first (And also include span names that we already have).

Another option is to define and adhere to the names of the metrics through this spec and adhere to that.

beeme1mr:
I think we need an additional dimension that represents the configuration of the feature flag being evaluated. The terminology varies depending on the implementation, but many have concepts like project, workspace, namespace, or app. Within that collection, there tend to be environment-specific configurations (e.g. dev, hardening, production). Depending on the provider, you would want to split impressions based on those concepts. Perhaps we could call it a `scope`, and it would be up to the provider to decide what the scope looks like. This currently isn't covered by the OTel spec but should be included if we can define what the property should be called.

Kavindu-Dodan:
> I think we need an additional dimension that represents the configuration of the feature flag being evaluated. The terminology varies depending on the implementation, but many have concepts like project, workspace, namespace, or app. Within that collection, there tend to be environment-specific configurations (e.g. dev, hardening, production). Depending on the provider, you would want to split impressions based on those concepts. Perhaps we could call it a `scope`, and it would be up to the provider to decide what the scope looks like. This currently isn't covered by the OTel spec but should be included if we can define what the property should be called.

Isn't this already covered by Evaluation Context [1], which is available through hook context [2] ? If so, I think we can include this as a dimension of `evaluationRequests` metric. 

[1] - https://openfeature.dev/specification/sections/evaluation-context
[2] - https://openfeature.dev/specification/sections/hooks/#41-hook-context

OFEP-metric-hooks.md
```diff
@@ -0,0 +1,67 @@
...
+
+A counter[1] based metric to calculate the number of successful flag evaluations. This is recorded at `after` stage of 
+the hook.
+
+### evaluationError
+
+A counter[1] based metric to calculate the number of failed flag evaluations. This is recorded at `error` stage of
+the hook.
+
+### evaluationActive
```
justinabrahms, line 33: Evaluation latency metric as well?

beeme1mr: I would imagine that evaluation latency should be extremely low, but I suppose it would depend on the provider. It certainly could be an interesting and useful metric. Perhaps the hook could allow you to opt-in or out of specific metrics.

beeme1mr:
> Isn't this already covered by Evaluation Context [1], which is available through hook context [2] ? Probably the intended identifier maps to the `targeting key` [3] 🤔

Yes, capturing that information as flag metadata is possible but I'm wondering if we can figure out what we want to call that property.

weyert:
Nice, I am having some metrics: https://github.com/Tapico/node-posthog-openfeature/blob/main/src/hooks/OpenTelemetryHook.ts

Works pretty well

OFEP-metric-hooks.md
Kavindu-Dodan: @justinabrahms  good point. I considered this option but I think this is not possible as hooks have a global presence compared to for example a HTTP request (request/response roundtrip calculation). Hooks have access to context details, but there is no way to uniquely separate an identifier between stages (for latency this is needed between the `before` & `after` stage). 

For example, if we consider the flag key, there could be parallel flag evaluations for the same flag. In Go, there is the `context` that can be used as a carrier for evaluation start time which is shared with all stages, but this is language specific and could be challenging to implement in other languages.

This is why I opted for `evaluationActive` which is simple and contains the number of active flag evaluations with dimensions for better slicing. 

Anyway, I am open to other opinion and let me know if you see any other way to approach this :)

Kavindu-Dodan:
> > Isn't this already covered by Evaluation Context [1], which is available through hook context [2] ? Probably the intended identifier maps to the `targeting key` [3] 🤔
> 
> Yes, capturing that information as flag metadata is possible but I'm wondering if we can figure out what we want to call that property.

`scope` is a good enough naming. We can also use something like`evaluation_context`or `targeting_key. 

Anyway, we should consider about privacy into account when using context data for metric dimensions, as these identifiers may include personal information such as user emails.

Kavindu-Dodan:
> Nice, I am having some metrics: https://github.com/Tapico/node-posthog-openfeature/blob/main/src/hooks/OpenTelemetryHook.ts
> 
> Works pretty well

That's nice :) If OFEP goes well, we will soon have official OpenFeature metrics hook for JS

OFEP-metric-hooks.md
beeme1mr: The sdk could add a time stamp on the evaluation context object and the hook could calculate a diff from that. I'm not sure if we would want to do that but it seems like an option.

toddbaert:
> > Isn't this already covered by Evaluation Context [1], which is available through hook context [2] ? Probably the intended identifier maps to the `targeting key` [3] 🤔
> 
> Yes, capturing that information as flag metadata is possible but I'm wondering if we can figure out what we want to call that property.

`scope` is a good enough naming. We can also use something like`evaluation_context`or `targeting_key. 

Anyway, we should consider about privacy into account when using context data for metric dimensions, as these identifiers may include personal information such as user emails.

`scope` is something I think we may want to consider bringing into our glossary. I think we should loosely define it as a "namespace" for flags. It could be used to logically segment environment, project, tenant, etc, since it's very possible to have flag name collisions between these.

OFEP-metric-hooks.md
```diff
@@ -0,0 +1,67 @@
...
+
+The proposal here is to define a set of metrics that can be used with different hook stages. For example, error metrics 
+can be collected at the hook's `error` stage. Going with this background, I propose the following metrics.
+
+### evaluationRequests
+
+A counter[1] based metric to calculate the number of flag evaluation requests. This is recorded at `before` stage of 
+the hook.
+
+### evaluationSuccess
```
toddbaert, line 23: I think it would help to mention some practical examples for some of these metrics. I think for `evaluationSuccess` in particular, they are basically what vendors refer to as "impressions". You might want to mention some such use-cases.

Not required for my approval though.

OFEP-metric-hooks.md
```diff
@@ -0,0 +1,67 @@
...
+
+A counter[1] based metric to calculate the number of successful flag evaluations. This is recorded at `after` stage of 
+the hook.
```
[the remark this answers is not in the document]
Kavindu-Dodan: Agree, I will include dimension details to each metric proposed through this OFEP

OFEP-metric-hooks.md
```diff
@@ -0,0 +1,67 @@
+## OFEP-hook-for-metrics
+
+## State: PENDING REVIEW
+
+This OFEP propose to introduce OpenFeature hook for OpenTelemetry metrics.
+
+## Background
+
+We already have OpenTelemetry Span support through hooks. Similarly, we can provide a dedicated hook for 
+OpenTelemetry
```
beeme1mr, line 10: Why is this on its own line?

OFEP-metric-hooks.md
```diff
@@ -0,0 +1,67 @@
...
+
+## State: PENDING REVIEW
+
+This OFEP propose to introduce OpenFeature hook for OpenTelemetry metrics.
+
+## Background
+
+We already have OpenTelemetry Span support through hooks. Similarly, we can provide a dedicated hook for 
+OpenTelemetry 
+metrics. Providing telemetry data out of the box will make OpenFeature attractive for both users and vendors.
```
beeme1mr, line 11:
```suggestion
metrics. Providing telemetry data out of the box will make OpenFeature attractive for both users and vendors.
```

OFEP-metric-hooks.md
```diff
@@ -0,0 +1,67 @@
...
+We already have OpenTelemetry Span support through hooks. Similarly, we can provide a dedicated hook for 
+OpenTelemetry 
+metrics. Providing telemetry data out of the box will make OpenFeature attractive for both users and vendors. 
+
+## Proposal
+
+The proposal here is to define a set of metrics that can be used with different hook stages. For example, error metrics 
+can be collected at the hook's `error` stage. Going with this background, I propose the following metrics.
+
+### evaluationRequests
```
beeme1mr, line 18: Could you derive this metric by aggregating successes and failures?

beeme1mr (commented):
Please review the metric names themselves. It looks like OpenTelemetry has naming guidelines that we should follow.

https://github.com/open-telemetry/oteps/blob/main/text/metrics/0108-naming-guidelines.md

beeme1mr (commented):
Please review the metric names themselves. It looks like OpenTelemetry has naming guidelines that we should follow.

https://github.com/open-telemetry/oteps/blob/main/text/metrics/0108-naming-guidelines.md

beeme1mr (commented):
Please review the metric names themselves. It looks like OpenTelemetry has naming guidelines that we should follow.

https://github.com/open-telemetry/oteps/blob/main/text/metrics/0108-naming-guidelines.md

OFEP-metric-hooks.md
Kavindu-Dodan: done

OFEP-metric-hooks.md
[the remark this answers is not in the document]
Kavindu-Dodan: Added common dimensions section and included unique dimention(s) of each metric we define through this OFEP

OFEP-metric-hooks.md
Kavindu-Dodan: Added a description to highlight usage of the metrics

OFEP-metric-hooks.md
Kavindu-Dodan: Yes, that's correct. evaluationRequests = evaluationSuccess + evaluationError

Note - I have updated the naming of metrics with latest commits

Kavindu-Dodan:
@thisthat @beeme1mr I updated metrics names proposed through this OFEP. They now have `feature_flag` prefix similar to sematic convention of feature flags [1] and meaningful suffices. 

- feature_flag.evaluation_active_count
- feature_flag.evaluation_requests_total
- feature_flag.evaluation_success_total
- feature_flag.evaluation_error_total

Let me know your thoughts on this 

[1] - https://opentelemetry.io/docs/specs/otel/logs/semantic_conventions/feature-flags/

toddbaert:
> @thisthat @beeme1mr I updated metrics names proposed through this OFEP. They now have `feature_flag` prefix similar to sematic convention of feature flags [1] and meaningful suffices.
> 
> * feature_flag.evaluation_active_count
> * feature_flag.evaluation_requests_total
> * feature_flag.evaluation_success_total
> * feature_flag.evaluation_error_total
> 
> Let me know your thoughts on this
> 
> [1] - https://opentelemetry.io/docs/specs/otel/logs/semantic_conventions/feature-flags/

This looks good to me.

toddbaert (approved):
I'm personally of the opinion this is clear enough to merge, and can be a basis for creating issues in the various contrib repos.

I'd love to hear once more from @thisthat or @beeme1mr as well.

I also think creating a PR to our spec glossary to define `scope` would be nice.

OFEP-metric-hooks.md
```diff
@@ -0,0 +1,85 @@
+## OFEP-hook-for-metrics
+
+## State: PENDING REVIEW
```
toddbaert, line 3: @Kavindu-Dodan I will update this to `APPROVED` if/when we get the required approvals.

thisthat (approved):

OFEP-metric-hooks.md
Kavindu-Dodan: Thank you :) let's wait for more feedback/reviews

beeme1mr (commented):
You should include a section on `scope` as a possible dimension. I think that will be required to use these metrics in a vendor-agnostic way. Once that's in place, I'm happy with the OFEP.

I'll mention this in the community meeting. If we have the approvals and no blockers, let's aim for merging this on Friday.

beeme1mr (commented):
You should include a section on `scope` as a possible dimension. I think that will be required to use these metrics in a vendor-agnostic way. Once that's in place, I'm happy with the OFEP.

I'll mention this in the community meeting. If we have the approvals and no blockers, let's aim for merging this on Friday.

OFEP-metric-hooks.md
```diff
@@ -0,0 +1,112 @@
...
+
+### Scope dimension
+
+credits - Michael Beemer
+
+It is possible to add an additional dimension representing the configuration of the feature flag being evaluated. A
+feature flag usually has a scope such as a project, workspace, namespace, or application. This can be further
+expanded to environment-specific configurations such as dev, hardening, and production or the cloud provider such as
+AWS, Azure or GCP. Adding this dimension through an agreed attribute name (suggested name - `scope`) benefits metric
+evaluations (ex:- drill down to cloud provider specific flag evaluations)
```
Kavindu-Dodan, line 83: @beeme1mr I credited your idea and added this to the expansion option

OFEP-metric-hooks.md
Kavindu-Dodan: I think I will resolve this with this change [1] and we can revisit this once we have the first implemention :) 

[1] - https://github.com/open-feature/ofep/pull/62/files#diff-1ee0937abd11c52172d0be2b854ab5162b15262da53e7ddbcd4090484879bc67R66

OFEP-metric-hooks.md
```diff
@@ -0,0 +1,112 @@
...
+### Metric to measure latency
+
+credits - Justin Abrahms
+
+Flag evaluation latency can be calculated with time measurements between `finally` and `before` stages. However,
+this requires time measurement to be shared between two stages, which require either a context propagation or a shared
+variable (potentially a map). Alternatively, SDK could mark the evaluation start timestamp to enhance the accuracy
+of the measurement
+
+### Scope dimension
```
beeme1mr, line 75: This needs to be included in the OFEP. It would be optional and only applied if a provider supplies it as flag metadata. Without it, these metrics are significantly less useful with providers that support multiple environments.

Kavindu-Dodan: I will first update go-sdk implementation & then update the OFEP

Kavindu-Dodan: I added a new section on `extra dimension` [1], backed by go-sdk implementation to prove the feasibility [2]

[1] - https://github.com/open-feature/ofep/pull/62/files#diff-1ee0937abd11c52172d0be2b854ab5162b15262da53e7ddbcd4090484879bc67R61-R85
[2] - https://github.com/open-feature/go-sdk-contrib/pull/217

beeme1mr (approved):

toddbaert:
I think we have sufficient consensus for this one. This is not a spec change, it just establishes some consistency around some of the OTel-related extensions/hooks. Merging.

OFEP-metric-hooks.md
```diff
@@ -0,0 +1,134 @@
+## OFEP-hook-for-metrics
+
+## State: PENDING REVIEW
```
toddbaert, line 3:
```suggestion
## State: APPROVED
```

PR #67, opened 2023-06-27
Flag Evaluation Wire Protocol

### LINKED ISSUE #65: [Proposal] Generic GRPC Provider
### Category of Proposal

Other

### Describe your proposal

# Context

In this [thread](https://cloud-native.slack.com/archives/C0344AANLA1/p1686835859830889), we've observed that it is challenging for open-source solutions to utilize Open Feature SDKs because there is no way to have lightweight client in a distribution that can communicate with every OpenFeature provider implementation.

This limitation compels these projects to support some form of plugins to enable users to add their desired providers.

While this approach can work with certain languages, it becomes more difficult with others _(such as C++, GO, etc.)_ and requires adapting these products to support the usage of the providers.

# Proposal

During the discussion, one idea was proposed: creating a generic GRPC provider capable of communicating with any vendor that implements the associated GRPC server for this provider.

Ideally, this generic GRPC provider should support both remote flag evaluation and local flag evaluation.

For local evaluation, we could request that the vendor also provide a standardized method for performing evaluation _(an option could be to use https://jsonlogic.com/ or a similar solution)_.

beeme1mr:
## This PR

- adds a flag evaluation wire protocol OFEP proposal

### Related Issues

Fixes #65


beeme1mr:
This OFEP potentially addresses @tcarrio [pluggable flagd OFEP](https://github.com/open-feature/ofep/pull/54) in an alternative way.

OFEP-flag-eval-wire-protocol.md
```diff
@@ -0,0 +1,54 @@
...
+- The schema contains references to flagd
+- The schema was not designed to be provider agnostic
+
+## Open source projects that may be interested in a wire protocol
+
+- OpenTelemetry demo application
+- Jenkins
+- Quarkus
+  - https://github.com/quarkusio/quarkus/issues/28032
+- Dapr
```
toddbaert, line 54: @oleg-nenashev @james-milligan if anyone has relevant discussions in other projects, we can link discussions here

OFEP-flag-eval-wire-protocol.md
```diff
@@ -0,0 +1,54 @@
...
+
+### Alternative 1: Use the existing flagd proto schema
+
+As an alternative to creating a new proto schema, we could reuse the existing flagd schema.
+This schema already complies with the OpenFeature specification and provides a working end-to-end implementation.
+
+The drawbacks to this alternative are:
+
+- The schema contains references to flagd
+- The schema was not designed to be provider agnostic
```
toddbaert, line 46: Another disadvantage is simply the name, which is not neutral, IMHO.

toddbaert (approved):
I think this is a compelling idea.

Defining such a wire protocol would mean applications could include vendor-neutral feature flags in a manner that "feels" similar to exposing Prometheus metrics, or the like.

OFEP-flag-eval-wire-protocol.md
beeme1mr: The first bullet point covers that.

OFEP-flag-eval-wire-protocol.md
```diff
@@ -0,0 +1,54 @@
...
+## Action items
+
+- Review the [existing proto schema](https://github.com/open-feature/schemas/blob/main/protobuf/schema/v1/schema.proto) to identify ways to make it provider agnostic.
+- Create a new proto schema that defines a general RPC-based flag evaluation based on the OpenFeature provider specification.
+- Publish the proto schema to the [Buf registry](https://buf.build/open-feature)
+- Create RPC providers for every OpenFeature SDK
+- Create flag evaluation server implementations
+  - flagd
+  - Go Feature Flag
+- Write documentation on how to create a new flag evaluation server.
```
thomaspoignant, line 32: Should we add an item about sharing this new integration point with the different vendors linked to open-feature?

thomaspoignant (commented):
Super nice 👌 
My guess is that it will ease the adoption of Open-Feature for new vendors interested by the project.

thomaspoignant (approved):

thisthat (approved):

OFEP-flag-eval-wire-protocol.md
beeme1mr: I added another action item for supporting vendors in the creation of a flag evaluation service. I think this would cover advertising and initial support.

toddbaert (approved):

OFEP-flag-eval-wire-protocol.md
```diff
@@ -0,0 +1,55 @@
...
+However, that's not always the case.
+For example, many open source applications could benefit from having a standardized means of including feature flags.
+A flag evaluation wire protocol would allow developers to choose the architecture that best fits their needs.
+
+## In-process vs Remote Evaluation
+
+Many feature flag vendors offer SDKs that perform flag evaluation in-process.
+This provides fast and consistent flag evaluations with no additional deployment complexities.
+Remote evaluation logic can be written in a single language and be used interchangeably with any system complying with the wire protocol.
+That makes it ideal for open source projects that need a vendor-agnostic solution for feature flags.
```
moredip, line 21: > That makes it ideal for open source projects that need a vendor-agnostic solution for feature flags.

This doesn't follow to me. A big part of the OpenFeature value prop is in providing a vendor-neutral (in-process) SDK, one which an open-source project can plug into without any awareness of what providers are plugged into that SDK.

OFEP-flag-eval-wire-protocol.md
```diff
@@ -0,0 +1,55 @@
...
+- Integrate with various open source projects looking to include feature flags.
+
+## Prior art and alternatives
+
+### Alternative 1: Use the existing flagd proto schema
+
+As an alternative to creating a new proto schema, we could reuse the existing flagd schema.
+This schema already complies with the OpenFeature specification and provides a working end-to-end implementation.
+
+The drawbacks to this alternative are:
```
moredip, line 44: If we don't use the existing flagd protocol then IMO we should plan on having flagd implement this new protocol. Delivering a protocol with no reference implementation doesn't seem great.

moredip (commented):
I'm really not sold on defining this protocol as part of OpenFeature. I think it distracts from the main focus of OF - the spec and SDK - and I'm skeptical that it will get vendor support. 

I'm not sure why we wouldn't just use the existing flagd protocol.

thomaspoignant (approved):

OFEP-flag-eval-wire-protocol.md
thomaspoignant: The problem that @oleg-nenashev reported with the current approach, is that it is hard for opensource projects to support all providers out of the box or to provide a plugin system for people to use a specific provider.

In some languages, we can do it easily (such as Java), but for some others such as Go it can be complex and will force these open-source projects to create a plugin system specific to open-feature to be able to support an unknown list of providers.

james-milligan (approved):

Kavindu-Dodan:
I think this is a good initiative and reduced complexity when OF needs to be adopted by other systems. And I tried to visualize how this would fit into our SDKs and given below is my understanding visualization in a diagram,

```mermaid
flowchart TD
    A[SDK] --> CTB(SDK Contrib)
    CTB --> VA(Vendor A) --> Backend(Vendor Server)
    CTB --> VB(Vendor B) --> ....
    A--> W(Wire Protocol)
    W --> IMPLA(Server A)
    W --> PR(Proxy A)
```

Left side (in-process)

- Current _SDK_ relly on _SDK Contrib_ to connect with the actual evaluation of the feature flag
-  _SDK Contib_  layer is sort of a trasformer. It transforms vendor-specific feature flags (most of the time) coming from vendor backend 
- Can the current _SDK_ to _SDK Contrib_ bound solution adopted by a generic solution (Jenkins, K8s) ? Yes, but then that solution needs to expose configurations to setup _SDK Contib_ dependencies 

Right side (wire protocol)

- Wire protocol is standardized, hence SDK will have built-in support
- Wire protocol directly connects to vendors or their proxy deployments
- Adoptions consume flag evaluations exposed from language-specific SDK based on the wire protocol

Besides, I think both approaches will live side-by-side through OF SDK as they have their own benefits.

Kavindu-Dodan (approved):

Kavindu-Dodan:
> If you're just a consumer of the platform then the feature flag ruleset is really just going to be another type of configuration - you're not going to need dynamic evaluation or anything like that

I see this differently. Configurations are a good start and this is the minimal setup of feature flags. But if we think about platform integrations, there is a need for dynamic configurability. This is the missing piece I see wire protocol can fulfill  

> It seems much more likely that Jenkins would just pick a specific fixed provider and ask its users to use that framework if they want to manage the flags

Wire protocol will avoid vendor locking, which is favorable for commercial/open-source platforms. Plus, this gives flexibility to adoptions so that they can use any feature flag vendor that supports the standard (more or less similar to OF SDKs)

> AND they'd now be requiring their consumers to run a new standalone process for flagging (very likely flagd)

I don't think this is mandatory for everyone. Default values will be the standard operating behavior. Flagging service will be only needed for advanced use cases such as platform offerings (ex:- Jenkins as a service, K8s as a service). Since wire protocol is standardized, this can be a custom or vendor-backed implementation. 

> ... broadens the entire scope of the OpenFeature to include this wire protocol, which is more confusing for new people coming to OpenFeature

Agree on the complexity -  we will have to maintain the wire protocol and provide a reference implementation(flagd). But I don't think this will confuse newcomers to feature flagging. SDK will continue to support conventional in-process evaluations [1]. Rather than confusing, I think this will enable more use cases. 

[1] - https://github.com/open-feature/ofep/pull/67#issuecomment-1615204626

toddbaert:
I don't believe we need 100% consensus in all things, however @moredip has been key in the project since it's inception, and this does represent a sort of new direction for OpenFeature.

@oleg-nenashev Do you have any thoughts in response to @moredip 's comments [here](https://github.com/open-feature/ofep/pull/67#issuecomment-1618837529)? And @Kavindu-Dodan 's responses [here](https://github.com/open-feature/ofep/pull/67#issuecomment-1618961154)? I think these are both full of good points.

Kavindu-Dodan:
Some points worth cross-referencing from the community meeting on 6th July 2023 [1] [2]

- Using flagd itself as the proxy 

flagd at the current state is a proxy exposing flags from different data sources. And this will require flagd to be a core component of each SDK and write convertors from other providers to flagd

-  caching & security

Client libraries heavily rely on caching to give flag evaluation performance. Wire protocol may have potential performance issues as it requires a proxy/server request. Further, caching must not be a concern of wire protocol implementation as it could lead to other issues, hence it should be minimal. 

Besides, we might also need to consider security for the SDK to proxy/server communication. This could be outsourced to a proxy but may be protocol will need to define a recommendation.

[1] - https://docs.google.com/document/d/1pp6t2giTcdEdVAri_2B1Z6Mv8mHhvtZT1AmkPV9K7xQ/edit#bookmark=id.9ltvj8qmp30v
[2] - https://docs.google.com/document/d/1pp6t2giTcdEdVAri_2B1Z6Mv8mHhvtZT1AmkPV9K7xQ/edit#bookmark=id.mp7ce32kaa5m

OFEP-flag-eval-wire-protocol.md
```diff
@@ -0,0 +1,55 @@
...
+## In-process vs Remote Evaluation
+
+Many feature flag vendors offer SDKs that perform flag evaluation in-process.
+This provides fast and consistent flag evaluations with no additional deployment complexities.
+Remote evaluation logic can be written in a single language and be used interchangeably with any system complying with the wire protocol.
+That makes it ideal for open source projects that need a vendor-agnostic solution for feature flags.
+
+## Action items
+
+- Review the [existing proto schema](https://github.com/open-feature/schemas/blob/main/protobuf/schema/v1/schema.proto) to identify ways to make it provider agnostic.
```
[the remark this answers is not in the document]
tcarrio: My immediate reaction to gRPC **only** was somewhat negative, since this restricts implementations somewhat. Even though clients can get flexibility with Web gRPC it's not shared in the server-side.

As a **wire protocol** I'm not entirely opposed.

But I would be interested in seeing OpenAPI definitions as part of this as well, unless there are significant issues with that approach.

justinabrahms:
Sorry for my delay. I'm getting lost on this proposal w/r/t jenkins and how they need it. It seems like we're talking about replacing jenkins' existing configuration mechanism with open feature. I'm not convinced that's a targeted use-case.

That said, what the proposal is reading to me as is: "Hey, it sucks to have to write N different providers where N is the number of languages that I, $vendor, would like to support." Instead, we're offering them a simpler "implement this interface over http and we'll support you in basically every language that we write SDKs in. **That** feels like a valuable thing.  Thoughts @moredip ?

beeme1mr:
Hey @justinabrahms, yes, I agree that's the immediate value. At the very least, it would consolidate the dev effort for Go Feature Flagd and flagd providers.

I included the Jenkins parts because @oleg-nenashev, a Jenkins maintainer, expressed interest in this functionality.

oleg-nenashev:
I am yet to review it in details, but I definitely express interest. Less for Jenkins, more for the ecosystem of small Golang/Rust projects that need to bundle something vendor neutral

kinyoklion (approved):

OFEP-flag-eval-wire-protocol.md
```diff
@@ -0,0 +1,55 @@
+## Flag Evaluation Wire Protocol
+
+## State: DRAFTING
+
+Create a wire protocol for RPC-based flag evaluation compliant with the OpenFeature specification.
```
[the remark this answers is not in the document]
beeme1mr: I could loosen the requirement to not specify rpc. The important part to me is to have a single source of truth like a protobuf or OpenAPI that can be used to generate clients. It's also important to have support for bi-directional communication for events and caching.

AloisReitbauer (approved):

toddbaert:
@beeme1mr should we close this with the addition of the OFREP to the spec repo and it's own repo? Is the OFEP still needed for record-keeping purposes you think?

I think https://github.com/open-feature/protocol/blob/main/README.md may accomplish the same goals :shrug:

thomaspoignant:
I think we can close this one, since [open-feature/protocol](https://github.com/open-feature/protocol) is the place where things are happening.

To follow the evolution of the project please follow to the protocol repository.

PR #71, opened 2023-07-06
feat: improving e2e test strategy

Kavindu-Dodan:
## This PR

Propose to improve OpenFeature SDK e2e test strategy by decoupling tests from flagd and introducing a self-contained provider built into SDK implementation 

Feel free to comment, propose or approve the ofep based on your experience

----
Note - While this OFEP is being reviewed, I will attempt to create a POC implementation of the proposed provider in of the SDK I maintain.


OFEP-sdk-e2e-test-strategy.md
```diff
@@ -0,0 +1,56 @@
...
+To fulfill these needs, I am proposing a simple in-memory provider to be built into the SDK and maintained alongside the SDK implementation. 
+
+## In-memory provider
+
+I am proposing following features for the in-memory provider,
+
+- Provider should be initiated with a pre-defined set of flags
+- Flag structure definition will be based on test requirements and must be minimal
+- EvaluationContext support for at least few predefined parameters
+- Must provider SDK contract implementations where needed (ex:- consider `NoOpProvider` [7])
```
thisthat, line 45: I suggest having an internal package used for testing, similar to what OTel has:
- Java: https://github.com/open-telemetry/opentelemetry-java/blob/aab71593270111ad3bd01e5e1c44a2e50aacf2c5/sdk/testing/src/main/java/io/opentelemetry/sdk/testing/exporter/InMemorySpanExporter.java
- Go: https://github.com/open-telemetry/opentelemetry-go/blob/c404a30b96f6117ac677c5e15119a4d724e2ab3e/sdk/trace/tracetest/exporter.go

thisthat (commented):
I fully support the idea. Beside the benefits for our own SDKs, this also should enhance DevExp providing some nice way for developers to facilitate the testing of their FF logic

OFEP-sdk-e2e-test-strategy.md
```diff
@@ -0,0 +1,56 @@
...
+- Tests must verify error scenarios 
+
+To fulfill these needs, I am proposing a simple in-memory provider to be built into the SDK and maintained alongside the SDK implementation. 
+
+## In-memory provider
+
+I am proposing following features for the in-memory provider,
+
+- Provider should be initiated with a pre-defined set of flags
+- Flag structure definition will be based on test requirements and must be minimal
```
toddbaert, line 43: I'm not sure about a set of pre-defined flags.

I think I'd rather have a constructor were you could pass a simple map of flags. In addition to "hard" flag values, some map entries could be lambdas that accept a context and return some flag value (to test context parsing/passing of the SDK under test).

OFEP-sdk-e2e-test-strategy.md
toddbaert, line 45: I think we need to add support for events too (init and shutdown) - some simple mechanism to update the flag config as well, which would send flag change events.

Soon we will have another gherkin test suite for events (see [here](https://github.com/open-feature/test-harness/issues/50)).

toddbaert:
I dont want us to build a "flagd" in each SDK - but I think what you're proposing (a simple in memory provider) per SDK makes sense. Indeed, as @thisthat points out, similar things have been implemented by OTel (such as the in-memory metrics exporters) and they fulfill similar roles (unit testing, etc).

I have highlighted a few additional requirements I think would be necessary to make this fully usable. I hope they make sense.

OFEP-sdk-e2e-test-strategy.md
Kavindu-Dodan: Yes, this is what I meant - the provider is constructed with a set of flags. I will update the description to emphasize this. 

💯  agree on context handling with lambda/callbacks - great idea to minimalize internals and let the caller control the context handling,

OFEP-sdk-e2e-test-strategy.md
Kavindu-Dodan: Yeah agree, I will include this point/sub point

Kavindu-Dodan: Thanks for the references. Yeah, packaging should highlight  the fact that these are test related

Kavindu-Dodan:
@thisthat & @toddbaert  thank you for the feedback

I have implemented a POC implementation of the proposed InMemory provider in go-sdk - https://github.com/open-feature/go-sdk/pull/192

toddbaert:
> @thisthat & @toddbaert thank you for the feedback
> 
> I have implemented a POC implementation of the proposed InMemory provider in go-sdk - [open-feature/go-sdk#192](https://github.com/open-feature/go-sdk/pull/192)

After looking at @Kavindu-Dodan 's go implementation, I agree we should do this in all SDKs.

toddbaert (approved):

OFEP-sdk-e2e-test-strategy.md
Kavindu-Dodan: Went with this approach - https://github.com/Kavindu-Dodan/go-sdk/tree/feat/in-memory-provider-for-testing/pkg/openfeature/testing :)

Kavindu-Dodan: I think this is needed by default. For the initial implementation, we can aim to replace flagd with in-memory setup. Later on, we can build the spec compliance tests

toddbaert:
@Kavindu-Dodan I think it's a good idea to actually add the exported in-memory provider as a spec requirement. What do you think? I think it would be a good testing utility even for consumers. We could add it as a new section, or maybe some kind of appendix.

OFEP-sdk-e2e-test-strategy.md
Kavindu-Dodan: Proposed features set expanded - https://github.com/open-feature/ofep/pull/71/files#diff-6ab079685fe6f42602ada091a5c5e3f40ec770e26488b8d33373f9d4084a3497R40

Kavindu-Dodan:
> @Kavindu-Dodan I think it's a good idea to actually add the exported in-memory provider as a spec requirement. What do you think? I think it would be a good testing utility even for consumers. We could add it as a new section, or maybe some kind of appendix.

I think this is a good idea. Adding an appendix section with extra details and recommendations will benefit the community

federicobond:
> After looking at @Kavindu-Dodan 's go implementation, I agree we should do this in all SDKs.

Agree too. The only thing I'm not convinced about is whether flags for the in-memory provider should be set statically at initialization only or a mutable interface would be more helpful in practice.

For example, it might be helpful for some tests to allow the provider to change the resolution of a flag without having to set a whole new provider. Think cascading flag overrides for tests, where a flag is set for a whole test class and another at the method level, while keeping the previous one.

Kavindu-Dodan:
> > After looking at @Kavindu-Dodan 's go implementation, I agree we should do this in all SDKs.
> 
> Agree too. The only thing I'm not convinced about is whether flags for the in-memory provider should be set statically at initialization only or a mutable interface would be more helpful in practice.
> 
> For example, it might be helpful for some tests to allow the provider to change the resolution of a flag without having to set a whole new provider. Think cascading flag overrides for tests, where a flag is set for a whole test class and another at the method level, while keeping the previous one.

I think the in-memory provider can be easily expanded if we see the need. For example, the provider constructor can accept flag storage that can be mutated independently. However, I am not sure whether we must define this through OFEP itself 🤔

OFEP-sdk-e2e-test-strategy.md
```diff
@@ -0,0 +1,61 @@
...
+For testing of the SDKs, current approach uses gherkin step definitions [2] contained in OpenFeature/test-harness [3] repository.
+And the test implementations use a real instance of flagd [4] to validate test rules.
+
+The current approach however comes with a problem. As SDK tests rely on flagd, these tests rely on SDK contributions [5].
+This means,
+
+- SDK cannot introduce tests till flagd contribution is there
+- Breaking changes require disabling e2e tests till
+  - Change propagate to respective SKD contribution repository
+  - flagd implements the breaking change
```
justinabrahms, line 20:
```suggestion
- Breaking changes require disabling e2e tests until
  - Changes propagate to respective SDK contribution repository
  - flagd implements the breaking change
```

OFEP-sdk-e2e-test-strategy.md
```diff
@@ -0,0 +1,61 @@
...
+harder to maintain in the long run.
+
+## Proposal
+
+Given OpenFeature has a lot of community interest and SDK, I am proposing to introduce a simplified e2e test strategy.
+The test strategy must comply with the following,
+
+- Tests must not use mocks and must run against an OpenFeature compliant provider
+- Tests must be self-contained except for test definitions which are common to all SDKs
+- Tests must verify error scenarios
```
justinabrahms, line 34:
```suggestion
Given OpenFeature has a lot of community interest and SDK, I am proposing to introduce a simplified e2e test strategy.
Gherkin tests must:
- not use mocks and must run against an OpenFeature compliant provider
- be self-contained except for test definitions which are common to all SDKs
- verify error scenarios 
```

I disagree with the wording of making this all tests, but if you're specifically referring to the gherkin tests, this seems fine with me.

OFEP-sdk-e2e-test-strategy.md
```diff
@@ -0,0 +1,61 @@
...
+To fulfill these needs, I am proposing a simple in-memory provider to be built into the SDK and maintained alongside the SDK implementation. 
+
+## In-memory provider
+
+I am proposing following features for the in-memory provider,
+
+- Flag structure definition will be based on test requirements and must be minimal
+- Provider is initiated with a pre-defined set of flags provided to constructor
+- EvaluationContext support should be provided through callbacks/lambda expressions to minimize implementation
+- Must provider SDK contract implementations where needed (ex:- consider `NoOpProvider` [7])
```
justinabrahms, line 45:
```suggestion
- Must provide SDK contract implementations where needed (ex:- consider `NoOpProvider` [7])
```

justinabrahms (changes requested):

OFEP-sdk-e2e-test-strategy.md
```diff
@@ -0,0 +1,61 @@
...
+
+SDKs are at the core of OpenFeature as they contain language specific implementations of the OpenFeature specification [1]
+
+For testing of the SDKs, current approach uses gherkin step definitions [2] contained in OpenFeature/test-harness [3] repository.
+And the test implementations use a real instance of flagd [4] to validate test rules.
+
+The current approach however comes with a problem. As SDK tests rely on flagd, these tests rely on SDK contributions [5].
+This means,
+
+- SDK cannot introduce tests till flagd contribution is there
```
toddbaert, line 17:
```suggestion
- SDK cannot introduce tests until flagd contribution is there
```

OFEP-sdk-e2e-test-strategy.md
```diff
@@ -0,0 +1,61 @@
...
+- Breaking changes require disabling e2e tests till
```
toddbaert, line 18:
```suggestion
- Breaking changes require disabling e2e tests until
```

OFEP-sdk-e2e-test-strategy.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Background
+
+SDKs are at the core of OpenFeature as they contain language specific implementations of the OpenFeature specification [1]
+
+For testing of the SDKs, current approach uses gherkin step definitions [2] contained in OpenFeature/test-harness [3] repository.
+And the test implementations use a real instance of flagd [4] to validate test rules.
+
+The current approach however comes with a problem. As SDK tests rely on flagd, these tests rely on SDK contributions [5].
+This means,
```
beeme1mr, line 15:
```suggestion
This means:
```

OFEP-sdk-e2e-test-strategy.md
```diff
@@ -0,0 +1,61 @@
...
+  - flagd implements the breaking change
+- Cannot write tests till
+  - flagd implements the feature
+
+Besides, there is a circular dependency among `Go SDK` - `Go SDK Contribution` - `flagd` [6] which will become 
+harder to maintain in the long run.
+
+## Proposal
+
+Given OpenFeature has a lot of community interest and SDK, I am proposing to introduce a simplified e2e test strategy.
```
beeme1mr, line 29:
```suggestion
Given OpenFeature has a lot of community interest and SDKs, I am proposing to introduce a simplified e2e test strategy.
```

OFEP-sdk-e2e-test-strategy.md
```diff
@@ -0,0 +1,61 @@
...
+The test strategy must comply with the following,
```
beeme1mr, line 30:
```suggestion
The test strategy must comply with the following:
```

beeme1mr:
I wonder if we could use this to automagically build out a support feature matrix and update a readme. Basically, if there are e2e tests for events, it's supported by the SDK. Definitely out of scope of this OFEP, but the idea may be worth exploring.

OFEP-sdk-e2e-test-strategy.md
Kavindu-Dodan: @justinabrahms Good point 🤔  my original intention of the OFEP was on e2e tests (Gherkin test suite). This is why I packaged the POC implementation of Go under `testing` [1]. And this closely follows the OTel testing framework  discussed here [2]

I am open to opinions here.

[1] - https://github.com/Kavindu-Dodan/go-sdk/tree/feat/in-memory-provider-for-testing/pkg/openfeature/testing
[2] - https://github.com/open-feature/ofep/pull/71#discussion_r1255278514

Kavindu-Dodan:
> I wonder if we could use this to automagically build out a support feature matrix and update a readme. Basically, if there are e2e tests for events, it's supported by the SDK. Definitely out of scope of this OFEP, but the idea may be worth exploring.

If I have to do this, I would rather create an OpenFeature GitHub action 🤔  Reason is, there is a considerable effort to do this in each language (Check feature completenes -> Update readmen -> Create Git PR )

OFEP-sdk-e2e-test-strategy.md
justinabrahms: I'm +1 w/ my proposed change.

Kavindu-Dodan: Just to confirm, your proposal is to make the in-memory provider SDK a built-in provider (not only for testing but as a fully-fledged provider). I can reword the proposal if there is no objection from others

toddbaert: > Just to confirm, your proposal is to make the in-memory provider SDK a built-in provider (not only for testing but as a fully-fledged provider). I can reword the proposal if there is no objection from others

I'm not against this. Obviously it's real-work usage will have a limited value, but I think it's fine to consider it a "real" provider, with more usage beyond testing.

tcarrio:
Just to clarify, does this mean that the in-memory provider will be **included** in the SDK, or part of the **repository**.

The latter would be my preference. This is something that we can accomplish without tying it directly into the same package and publishing the provider separately, but we do retain atomic changes.

That way no one has to unnecessarily install the in-memory provider code unless they actually wanted to use an in-memory provider (which I can't think of many cases that actually makes sense in production).

toddbaert:
> Just to clarify, does this mean that the in-memory provider will be **included** in the SDK, or part of the **repository**.
> 
> The latter would be my preference. This is something that we can accomplish without tying it directly into the same package and publishing the provider separately, but we do retain atomic changes.
> 
> That way no one has to unnecessarily install the in-memory provider code unless they actually wanted to use an in-memory provider (which I can't think of many cases that actually makes sense in production).

I see your point, and I'd like to come to a consensus on this before accepting/merging this OFEP.

I tend to think it would be better to actually include it in the SDK. Here are my reasons why:

- The provider will be useful to others, not just SDK authorsl. I think Integrations will leverage it for testing. If we don't ship it the SDK, but instead only keep it in the repo, this won't be possible, or we'll have to convert SDK repos into monorepos.
- The provider should be _very small_ from a code perspective, and have no dependencies. Shipping it with the SDK should not result in a significantly larger binary. Static linking (as in Go) and tree shaking (as in JS) can be leveraged to prevent its being included in consuming applications if we build and package it with that in mind.
- We currently don't separate the API which application authors target from the toolkit used to create integrations, like OTel does. We may do this in the future (by publishing a `OpenFeature API` package). I would put the in-memory provider in the category of "toolkit used to create integrations", so for now, I think it should be included.

OFEP-sdk-e2e-test-strategy.md
Kavindu-Dodan: Then I will reword the OFEP to highlight this fact. The in-memory provider will be bundled with SDK and not necessarily a test only provider.

Kavindu-Dodan: @justinabrahms & @toddbaert expanded the in-memory provider to be a general provider (not test only) [1]

[1] - https://github.com/open-feature/ofep/blob/ec965bcf29f7eab763f04a46c1cf19e1d55ba5f8/OFEP-sdk-e2e-test-strategy.md#in-memory-provider

justinabrahms (approved):

OFEP-sdk-e2e-test-strategy.md
```diff
@@ -0,0 +1,58 @@
+## OFEP: SDK end-to-end test strategy
+
+## State: Drafting
+
+This OFEP proposes to introduce a simplified process to write SDK end-to-end(e2e) tests
+
+## Background
+
+SDKs are at the core of OpenFeature as they contain language specific implementations of the OpenFeature specification [1]
```
beeme1mr, line 9: Please update the links to be footnotes. You can see the syntax here.

https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#footnotes

beeme1mr (approved):

OFEP-sdk-e2e-test-strategy.md
Kavindu-Dodan: This is cool - see the preview : https://github.com/open-feature/ofep/blob/2f7c045aebfe0c7ebed76354ac6f57628a91ad2c/OFEP-sdk-e2e-test-strategy.md#user-content-fn-2-b1b3f960f2d98fa55a3fcd378c954b6f

PR #80, opened 2023-08-03
SDK wait provider to be ready

thomaspoignant:
## This PR

- adds a proposal to be able to wait in the SDK for the provider to be ready.


OFEP-sdk-wait-provider-ready.md
```diff
@@ -0,0 +1,58 @@
...
+```
+
+## Proposal
+It would be great to have a way in the SDKs to wait for the provider to be ready.
+
+
+## Alternatives
+
+### Alternative 1: make `setProvider` synchronous
+Probably not the most ideal solution but we could change the behavior of `setProvider` to wait for the initialize function to throw or return.
```
kinyoklion, line 38: Depending on the language async is not always that great for waiting for something that you want to actually allow to complete asynchronously. (Java before completable futures, Rust where generally it will only progress if you are polling it.) Likely you end up making two methods anyway.

OFEP-sdk-wait-provider-ready.md
```diff
@@ -0,0 +1,58 @@
...
+It would be great to have a way in the SDKs to wait for the provider to be ready.
+
+
+## Alternatives
+
+### Alternative 1: make `setProvider` synchronous
+Probably not the most ideal solution but we could change the behavior of `setProvider` to wait for the initialize function to throw or return.  
+*It is a breaking change from the actual implementation.*
+
+### Alternative 2: Add a waiting function in the SDKs to wait for the provider to be ready
```
kinyoklion, line 41: I think this is reasonable. It provides a good deal of flexibility. I think, depending on language, you still end up also adding a timeout. Waiting forever isn't a great idea, and similar to the initial problem not all languages provide the ease of handling the timeout. (In JS you just race the promise with a time promise.)

`waitReady(timeout)`.

I do wonder if there is any value in difference here between core spec and stuff you can already build with the spec? It is inconvenient in java, but not impossible. Is there some other level we should have for things that can be implemented as a layer? I suppose if it is just an optional part of the spec that would be sufficient.

toddbaert: > I think this is reasonable. It provides a good deal of flexibility. I think, depending on language, you still end up also adding a timeout. Waiting forever isn't a great idea, and similar to the initial problem not all languages provide the ease of handling the timeout. (In JS you just race the promise with a time promise.)

Agreed 100%.

> I do wonder if there is any value in difference here between core spec and stuff you can already build with the spec? It is inconvenient in java, but not impossible. Is there some other level we should have for things that can be implemented as a layer? I suppose if it is just an optional part of the spec that would be sufficient.

I've been considering something like this as well, but I think an optional requirement (MAY or SHOULD) would be sufficient for this.

MAY or SHOULD, in combination with the nature of the requirement, strongly imply this would be a convenience method, which I believe is basically the case here.

toddbaert:
Should we recommend this functionality on the API, the client, or both?

Adding it just at the API seems like the most conservative move to me, especially if we're expecting it to throw. That would make this more of a convenience method for `integrators`. I'm not sure how I'd expect this to behave on clients, which we don't want to throw, generally.

OFEP-sdk-wait-provider-ready.md
```diff
@@ -0,0 +1,58 @@
...
+const client = Openfeature.getClient()
+await client.isReady()
+```
+
+### Alternative 2: Chain `waitReady()` with `setProvider(...)`
+Add a chain function to `setProvider(...)` to wait until the intialization is done.
+It will wait until the `initialize` function throw or return.
+
+```javascript
+Openfeature.setProvider(myprovider).waitReady()
```
beeme1mr, line 56:
```suggestion
await Openfeature.setProvider(myprovider).waitReady()
```

I'm assuming, in this case, the `waitReady` function would return a promise.

OFEP-sdk-wait-provider-ready.md
```diff
@@ -0,0 +1,58 @@
...
+
+```javascript
+Openfeature.setProvider(myprovider)
+const client = Openfeature.getClient()
+await client.isReady()
+```
+
+### Alternative 2: Chain `waitReady()` with `setProvider(...)`
+Add a chain function to `setProvider(...)` to wait until the intialization is done.
+It will wait until the `initialize` function throw or return.
```
beeme1mr, line 53:
```suggestion
It will wait for the `initialize` function to throw or return.
```

OFEP-sdk-wait-provider-ready.md
```diff
@@ -0,0 +1,58 @@
...
+Adding a waiting function will allow to block the SDK until the provider is ready.  
+This new function will wait for a `ready` event or a timelimit for the provider to be ready.  
+
+```javascript
+Openfeature.setProvider(myprovider)
+const client = Openfeature.getClient()
+await client.isReady()
+```
+
+### Alternative 2: Chain `waitReady()` with `setProvider(...)`
```
beeme1mr, line 51: I like that you can wait for a provider to be ready before a client has even been created. This helps keep a nice separation between the [application integrator](https://openfeature.dev/specification/glossary#application-integrator) and the [application author](https://openfeature.dev/specification/glossary#application-author).

beeme1mr (commented):
Thanks for putting this together, @thomaspoignant.

OFEP-sdk-wait-provider-ready.md
Kavindu-Dodan, line 51:
```suggestion
### Alternative 3: Chain `waitReady()` with `setProvider(...)`
```

Kavindu-Dodan, line 51: Chaining essentially require to change the signature (hence a breaking change) and adds complexity for SDK implementation.

I would rather expose a dedicated blocking _setProvider_ along with current async _setpProvider_ . For example, Java gRPC expose similar methods - `RouteGuideGrpc.newBlockingStub` vs `RouteGuideGrpc.newStub` [^1] 

[^1]: https://grpc.io/docs/languages/java/basics/#instantiating-a-stub

thomaspoignant: Should we add the option of using a blocking `setProvider` as an alternative in the OFEP?
How should we call this function? `setProviderSync`, `setProviderBlocking`, `setProviderAndWait` ?

thomaspoignant: Yes, I like the fact that we are waiting for the `provider` to be ready and not the `client` to be ready.

beeme1mr: Yeah, I think an alternative 4 would make sense. The name itself isn't that important at this stage of the proposal.

Kavindu-Dodan: +1 for alternative 4

OFEP-sdk-wait-provider-ready.md
```diff
@@ -0,0 +1,81 @@
...
+### Alternative 3: Chain `waitReady()` with `setProvider(...)`
+Add a chain function to `setProvider(...)` to wait until the intialization is done.
+It will wait for the `initialize` function to throw or return.
+
+```javascript
+await Openfeature.setProvider(myprovider).waitReady()
+const client = Openfeature.getClient()
+```
+
+### Alternative 4: Add a blocking `setProvider`
```
beeme1mr, line 60: This is my preferred option.

thomaspoignant: Same for me, I would go for option 4.

Kavindu-Dodan: Given that we have existing reference implementations on other frameworks (ex:- gRPC), I am also voting for this option. I think this is the easiest option with respect to implementation efforts.

Kavindu-Dodan (approved):

OFEP-sdk-wait-provider-ready.md
```diff
@@ -0,0 +1,81 @@
...
+    public void accept(EventDetails eventDetails) {
+        completableFuture.complete(eventDetails);
+    }
+});
+completableFuture.get();
+```
+
+## Proposal
+It would be great to have a way in the SDKs to wait for the provider to be ready.
+
```
beeme1mr, line 33: @thomaspoignant, could you please update the proposal section with an example from section 4? I think we could approve and merge after that.

thomaspoignant: I've added examples + change the status of the OFEP.

beeme1mr (approved):

toddbaert:
@thomaspoignant @beeme1mr My only remaining question about this proposal is [here](https://github.com/open-feature/ofep/pull/80#issuecomment-1664602950).

I think we should clearly indicate if we want this feature to be available only where providers are set (on the API) or also on the client. I advocate we add it just at the API level.

Kavindu-Dodan:
> @thomaspoignant @beeme1mr My only remaining question about this proposal is [here](https://github.com/open-feature/ofep/pull/80#issuecomment-1664602950).
> 
> I think we should clearly indicate if we want this feature to be available only where providers are set (on the API) or also on the client. I advocate we add it just at the API level.
> 
> In any case, when this is merged, I think we should add a `SHOULD` requirement to the spec for this.

I am not sure what you refer by `or also on the client` 🤔  we only have set provider in API level. And if set provider is blocking, then we cannot invoke API to retrieve client. So I don't see why we need to consider client here. But may be I am missing something.

toddbaert:
> I am not sure what you refer by `or also on the client`  we only have set provider in API level. And if set provider is blocking, then we cannot invoke API to retrieve client. So I don't see why we need to consider client here. But may be I am missing something.

`client.wairForReady` is what I meant, but I would rather not implement this.

toddbaert (approved):
Approved.

I recommend a new "SHOULD" requirement in the spec.

toddbaert:
Merging this. I think we see a trend toward option 4 as mentioned [here](https://github.com/open-feature/ofep/pull/80#issuecomment-1669106825). I think we should open a spec PR.

PR #85, opened 2023-08-31
Variations and Variables

fahad19:
## This PR

Introduces a new OFEP as per a previous discussion on [Slack](https://cloud-native.slack.com/archives/C0344AANLA1/p1692027588778849?thread_ts=1691952307.117719&cid=C0344AANLA1).

Proposes a new model for features, breaking it down into 3 kinds of values all scoped under a single feature:

- feature flag
- variation
- variables

## Notes

This is being proposed to help open up the possibility to create OpenFeature providers for Featurevisor: https://featurevisor.com/

SDK docs showing examples of the proposed feature model: 

- https://featurevisor.com/docs/sdks/

Further guides for the concepts introduced:

- https://featurevisor.com/docs/use-cases/experiments/
- https://featurevisor.com/docs/use-cases/remote-configuration/

---

I understand this proposal can be very different than your overall vision around feature flags, but I feel it's worth raising it to you all because it enables many other tools (both Open Source & SaaS) to create OpenFeature Providers then ❤️

### DIFF
--- a/012-OFEP-variations-and-variables.md
+++ b/012-OFEP-variations-and-variables.md
@@ -1,4 +1,4 @@
-## NAME
+## Variations and Variables
 
 ## State: DRAFTING
 
--- a/012-OFEP-variations-and-variables.md
+++ b/012-OFEP-variations-and-variables.md
@@ -1,0 +1,167 @@
+## NAME
+
+## State: DRAFTING
+
+Introduce two new kinds of evaluations under individual features called **variations** and **variables**, next to feature flag's own value. This will allow OpenFeature SDKs to support a wider range of feature management use cases with already existing tools.
+
+## Background
+
+The SDK API currently exposes various type-specific methods for evaluating a feature flag's single value:
+
+```js
+import { OpenFeature } from "@openfeature/web-sdk";
+
+OpenFeature.setProvider(new YourProviderOfChoice());
+
+const client = OpenFeature.getClient();
+
+// evaluation
+const isEnabled = client.getBooleanValue("my_feature", false);
+```
+
+The evaluated value could have been of other types as well, including:
+
+- `String`
+- `Number`
+- `Object`
+
+## Proposal
+
+This proposal aims to look at feature management as a broad topic, where flags are one of the implementations.
+
+OpenFeature, in its current form, is designed to support feature flags alone with their own single value.
+
+In real world implementations of various different tools (both Open Source and SaaS), we notice that the feature flags go beyond just a single value, even if that single value can be of various different primitive types for different use cases.
+
+This is where we introduce the concept of **Variations** and **Variables**.
+
+## Kinds of evaluations
+
+### Flags
+
+This is solely for checking if the feature flag itself is enabled or not, and is always of `Boolean` type.
+
+Example:
+
+```js
+const featureKey = "my_feature";
+const defaultState = false;
+
+const isEnabled = client.isEnabled(featureKey, defaultState);
+```
+
+### Variations
+
+We now enter the realm of A/B testing and experimentation.
+
+An experiment consists of different variations (which are usually of `String` type), and the user is consistently bucketed (assigned) into one of its variations, irrespective of the device or session the user is in.
+
+You can think of an experiment named `my_experiment`, that can have two variations:
+
+- `control`: the default behaviour
+- `treatment`: the new behaviour that we wish to test
+
+A user will be bucketed into one of these variations, and will stay in that variation for the duration of the experiment.
+
+Example:
+
+```js
+const featureKey = "my_experiment";
+const defaultVariation = "control";
+
+const variation = client.getVariation(featureKey, defaultVariation);
+```
+
+#### Weights
+
+Each variation is also associated with a weight, which is a number between 0 (0%) and 1 (100%), and represents the probability of the user being bucketed into that variation. The total weight of all variations in an experiment must be 1.
+
+This level of details is something to be handled at Provider level, and not the OpenFeature SDKs.
+
+#### Measuring success
+
+Applications will be tracking various different conversion goals (like successful user sign ups, purchases, etc). Based on the variation the user is bucketed into, we will be able to track the conversion goal events against each variation of the experiment over a period of time, and decide which variation is the winner.
+
+OpenFeature SDKs will not be involved here.
+
+### Variables
+
+Feature management can go beyond just on/off values for the flag itself and supporting a mechanism for a/b testing with variations.
+
+We may also want to support the concept of variables for supporting configuration to be deployed independently of our applications, which are values that can be changed over time.
+
+These values can be of various different primitive types for different use cases.
+
+You can think of a feature named `my_feature`, that can have several variables scoped under it:
+
+- `showSidebar`: boolean
+- `buttonColor`: string
+- `allowSignup`: boolean
+
+The variable values can be evaluated against the context that already exists in the SDK, and can also react further based on the bucketed variation if there's an experiment involved.
+
+Example:
+
+```js
+const featureKey = "my_feature";
+const variableKey = "showSidebar";
+const defaultValue = false;
+
+const showSidebar = client.getVariableBoolean(featureKey, variableKey, defaultValue);
+```
+
+## Summarizing the feature model
+
+Three different kinds of values under a single feature:
+
+- Flag status (`Boolean`)
+- Variation (`String`)
+- Variables:
+  - key (`String`)
+  - Value (`Boolean`, `String`, `Number`, `Object`)
+
+**Note**: It is worthwhile to expand the `Number` type into more explicit types like `Integer` and `Double`, especially when dealing with SDKs in strongly typed languages.
+
+## API directions
+
+With the new kinds of evaluations scoped under a single feature, we can consider the following API allowing more tools to create Providers targeting OpenFeature SDKs in various languages:
+
+```js
+import { OpenFeature } from "@openfeature/web-sdk";
+
+OpenFeature.setProvider(new YourProviderOfChoice({
+  // purely up to Providers to define this API
+  onActivation: () => console.log("handle experiment activation for tracking purposes")
+}));
+
+const client = OpenFeature.getClient();
+
+client.setContext({
+  userId: "123",
+  country: "nl",
+});
+
+const featureKey = "my_feature";
+const variableKey = "showSidebar";
+
+// feature flag's own status (always a boolean)
+const isEnabled = client.isEnabled(featureKey, false);
+
+// variation for a/b testing (always a string)
+const variation = client.getVariation(featureKey, "control");
+
+// variables (could be various different types)
+const showSidebar = client.getVariableBoolean(featureKey, variableKey, false);
+```
+
+## Additional notes
+
+The proposal is primarily motivated for integrating [Featurevisor](https://featurevisor.com) with OpenFeature SDKs, which is an open source Git-based feature management tool supporting the proposed model.
+
+If approved, this model can be used by other tools as well opening up OpenFeature's adoption to even a wider community.
+
+Regarding the proposed API, you can refer to documentation of Featurevisor SDK for getting a feel of real-world usage:
+
+- [`isEnabled`](https://featurevisor.com/docs/sdks/#checking-if-enabled)
+- [`getVariation`](https://featurevisor.com/docs/sdks/#getting-variations)
+- [`getVariable`](https://featurevisor.com/docs/sdks/#getting-variables)


012-OFEP-variations-and-variables.md
```diff
@@ -0,0 +1,167 @@
...
+The variable values can be evaluated against the context that already exists in the SDK, and can also react further based on the bucketed variation if there's an experiment involved.
+
+Example:
+
+```js
+const featureKey = "my_feature";
+const variableKey = "showSidebar";
+const defaultValue = false;
+
+const showSidebar = client.getVariableBoolean(featureKey, variableKey, defaultValue);
```
beeme1mr, line 110: I like the concept, especially with the type enforcement at the variable level. We can do something similar, but your proposal may be a nicer dev experience.

Here's what's already possible.

```ts
const featureKey = "my_feature";
const defaultValue = { showSidebar: false };

const { showSidebar } = client.getObjectValue(featureKey, defaultValue);
```

In TypeScript, it would even be possible to register a hook that validates the object before returning it to the client.

```ts
import { z } from "zod";

const myFeatureSchema= z.object({
  showSidebar: z.boolean(),
});
type MyFeature = z.infer<typeof myFeatureSchema>;

const featureKey = "my_feature";
const defaultValue: MyFeature = { showSidebar: false };

class Validator {
	constructor(private readonly zodSchema) {}

	after(_, evaluationDetails): {
		this.schema.parse(evaluationDetails.value);
	}
}

const { showSidebar } = client.getObjectValue<MyFeature >(featureKey, defaultValue, undefined, { hooks: [new Validator(myFeatureSchema)] });
```
> NOTE: This was written from memory and has yet to be tested.

The second example is fairly complex but would offer runtime variable validation.

012-OFEP-variations-and-variables.md
```diff
@@ -0,0 +1,167 @@
...
+
+Three different kinds of values under a single feature:
+
+- Flag status (`Boolean`)
+- Variation (`String`)
+- Variables:
+  - key (`String`)
+  - Value (`Boolean`, `String`, `Number`, `Object`)
+
+**Note**: It is worthwhile to expand the `Number` type into more explicit types like `Integer` and `Double`, especially when dealing with SDKs in strongly typed languages.
```
beeme1mr, line 123: This is already supported in strongly typed languages.

beeme1mr (commented):
Hey @fahad19, great proposal! Most of what you proposed is already technically possible but may benefit from more user-friendly method names. For reference, the method names were originally inspired by [this API comparison](https://github.com/open-feature/ofep/blob/main/prior-art/api-comparision.md).

012-OFEP-variations-and-variables.md
fahad19: `client.getObject()` can work, but this will require the provider to compute all the variables together and return it as an object, which application can destruct from.

the proposed API in document suggests we evaluate individual variables scoped under each feature on-demand.

also when the method name specifically says `getVariable*`, it shows the intent and the kind of desired value as well. We want the "variable" to be evaluated with that specific key.

012-OFEP-variations-and-variables.md
```diff
@@ -0,0 +1,167 @@
...
+
+This is solely for checking if the feature flag itself is enabled or not, and is always of `Boolean` type.
+
+Example:
+
+```js
+const featureKey = "my_feature";
+const defaultState = false;
+
+const isEnabled = client.isEnabled(featureKey, defaultState);
```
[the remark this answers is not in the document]
fahad19: if we consider features to contain a single value only, then yes it's a syntatic sugar.

but if we wish to establish three separate kinds of values, like:

- flag status (boolean)
- variation (string)
- variables (key/value pairs with different types of values)

then `.isEnabled()` ends up with a meaning of its own, which evaluates the flag's status as boolean.

012-OFEP-variations-and-variables.md
fahad19: I will remove this line soon.

012-OFEP-variations-and-variables.md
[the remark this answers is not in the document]
kinyoklion: I feel like it would be very difficult to have something meaningful here cross-vendor. This feels more like the "why" that would be associated with the reason or meta-data associated with a flag evaluation. (The detailed evaluation reason would be "on").

I don't think there is a consistent idea of what "on" would be, or if all vendors would have such a concept.

012-OFEP-variations-and-variables.md
kinyoklion: I think I had discussed this in another OFEP or spec comment, but I think that the idea of a "scoped" feature could be added and would probably be more widely applicable. 

Variables for this system seem closer to what variations may generally be from other systems (multiple types, potentially objects), and you would use several "flags" to accomplish something similar in those systems (or one structured flag containing the relevant data).

Here they are together in that a flag contains a few more states that are exposed.

If you were to imagine this at a very basic level it would be something like.

"myFeature" -> Variation (with on/off as details of that evaluation). Maybe this means some extra work is done.

"myFeature:showSideBar" -> Variable.

Being strings is for illustrative purpose, but if there is true need, then some more specific way could be devoted to support the idea.

Currently I think this is divergent enough that it would create a lot of dead API surface for most providers. There may be a way to make it into something more uniform though. How do we adjust the API so that two providers are not completely divergent from an API usage perspective?

toddbaert: > If you were to imagine this at a very basic level it would be something like.
> 
> "myFeature" -> Variation (with on/off as details of that evaluation). Maybe this means some extra work is done.
> 
> "myFeature:showSideBar" -> Variable.
> 
> Being strings is for illustrative purpose, but if there is true need, then some more specific way could be devoted to support the idea.

This is an interesting idea.

> Currently I think this is divergent enough that it would create a lot of dead API surface for most providers.

I haven't had time yet to really dig into this proposal, but I think as it stands, this is my biggest concern with it.

liran-p: >another challenge would be emitting an "activation" event when an Experiment has been exposed to end users.

There is PROVIDER_CONFIGURATION_CHANGED [event](https://github.com/open-feature/spec/blob/main/specification/sections/05-events.md#requirement-511) , you can use it with [event details](https://github.com/open-feature/spec/blob/main/specification/types.md#provider-event-details) to reflect it. Can this solve your challenge ?

fahad19: The challenge is primarily on the triggering side, not listening side.

For listening, I am expecting Provider can also accept an option like this:

```js
OpenFeature.setProvider(new FeaturevisorProvider({
  onActivation(featureKey, variationValue) {
    // do tracking here
  }
}));
```

It's the triggering part, which happens via Featurevisor SDK instance like this:

```js
const instance; // Featurevisor SDK instance

// .activate() is like .getVariation(), 
// but it also bubbles an event up for listening
const variation = instance.activate("myFeatureKey", context);
```

SDK docs for activation: https://featurevisor.com/docs/sdks/#activation

Based on the comment thread here, the evaluation of a feature's variation could happen like this with OpenFeature:

```js
const variation = client.getStringValue("myFeatureKey:variation", "control", context);
```

Which is equivalent to Featurevisor SDK's `.getVariation()` method: https://featurevisor.com/docs/sdks/#getting-variations

But what could be an idiomatic way of calling the `.activate()` method with OpenFeature?

fahad19: Internally for event management, it can potentially solve the case.

but which API method of OpenFeature client to use in this case? what API call from the OpenFeature SDK client instance should trigger the "activate" event, knowing that we aren't extending the client API with more methods besides existing `get*Value()` methods?

usually in a/b testing libraries, the activation event is intentionally triggered by the application when it is known that the user has been exposed to a certain experiment (feature in our case) once. we do not trigger it right when we evaluate the value. it's a matter of visibility here.

that's where I am struggling to fit the OpenFeature model here.

liran2000: Maybe by "activating" you are referring to flag configuration change, like other changes of changing flag values for example from disabled to enabled ?
As far as I know, the OpenFeature client is dealing with application reading the flags values, not editing the flags configuration, as editing flags configurations and values is done by the feature flags service (e.g. FeatureVisor).
Is that what you mean, or can you elaborate ?

fahad19: I will try to summarize the differences below:

## 1: configuration update

this is what happens when the fetched configuration (Featurevisor calls them datafiles) has new changes, and the evaluated values can change further.

links:

- https://featurevisor.com/docs/building-datafiles/
- https://featurevisor.com/docs/sdks/#refreshing-datafile
- https://featurevisor.com/docs/sdks/#events
- https://featurevisor.com/docs/integrations/partykit/

## 2: feature activations

this happens when we wish to track when a particular user has been bucketed into a specific variation of an experiment. 

we intentionally activate the feature/experiment from the client, it bubbles an event up to our listener, and then we send the event to a tracker of our choice (could be Google Analytics, SegmentIO, or anything else)

links

- https://featurevisor.com/docs/use-cases/experiments/
- https://featurevisor.com/docs/sdks/#activation
- https://featurevisor.com/docs/bucketing/

Activation does not change the configuration in anyway.

moredip:
my 2 cents is that if there's a way we can achieve this in "userland" (e.g. using "foo:bar" key names with semantics that the provider understands) then we should strongly prefer that over changing/extending the spec

012-OFEP-variations-and-variables.md
beeme1mr: You could bubble up events using hooks. A hook can be configured to run after each flag evaluation and even be included in the provider itself. It's also possible to pass "hook hints" at evaluation time. These can be used in a hook to customize the behavior even further.

fahad19: > A hook can be configured to run after each flag evaluation

this is what I am specifically trying to avoid. we should not emit an event at the time of evaluation.

the event should be emitted only when the application thinks the user has been exposed to that feature/experiment, manually.

toddbaert: > Hey just adding my voice to this as we @ DevCycle have a somewhat similar hierarchy to what's described here, having Features as our top-level organizational object that contain multiple variables (flags) underneath: https://docs.devcycle.com/introduction/core-concepts/feature-hierarchy https://docs.devcycle.com/sdk/server-side-sdks/node/node-openfeature https://github.com/DevCycleHQ/js-sdks/blob/main/sdk/openfeature-nodejs-provider/src/DevCycleProvider.ts
> 
> We generally have learned that the vast majority of folks want to just interface with flags (variables) as the main interface to our SDKs. There are very few folks who need the information about the feature + variation data at run-time, mostly that data is just needed for reporting / eventing. That is certainly something we've struggled with ourselves with OpenFeature is how to get the variation + feature metadata attached to the resolution details.
> 
> I think a better approach here would be to figure out how to keep the core interface of the OpenFeature spec the same, where users are interfacing with different types of flags (variables), but exposing the feature + variation data as part of the Resolution Details / [Detailed Evaluation](https://openfeature.dev/docs/reference/concepts/evaluation-api#detailed-evaluation) data.

This is an interesting perspective, and an appealing idea, to me at least.

@fahad19 do you think we could add well-defined properties to the [detailed evaluation](https://openfeature.dev/specification/types#evaluation-details) methods that would help? `value` would still maintain its current semantics (variable in DevCycle language as @jonathannorris mentions), but we could perhaps add additional properties that would satisfy your use-case, while remaining coherent with systems that don't have such concepts.

fahad19:
> my 2 cents is that if there's a way we can achieve this in "userland" (e.g. using "foo:bar" key names with semantics that the provider understands) then we should strongly prefer that over changing/extending the spec

@moredip: you are very much right in suggesting if we can a solution in the userland for scoped variations/variables like using `foo:bar` as keys.

I like this suggestion. and likely creating a new Featurevisor provider for OpenFeature following this pattern.

my response would be the same if I were an OpenFeature maintainer.

the only remaining thing for me is to figure out the "activation" event to be triggered at will, which is discussed in above thread.

012-OFEP-variations-and-variables.md
fahad19: thanks @toddbaert and @jonathannorris for all the inputs <3

I don't think the discussion here regarding activation will go anywhere, except for pushing it to the userland entirely.

I will close it with a summarized comment in a bit.

fahad19:
summary:

- for scoped variables, we can follow `featureKey:variableKey` pattern, keeping the OpenFeature API the same
- for experiment activation (which are not always handled at the time of evaluation, but when application decides when it has been exposed to the user visually), we can delegate that to the userland (application) without touching OpenFeature SDKs in any form

really appreciate all the inputs <3

I would have defended the OpenFeature API the same way to not have any further tweaks in the API to accommodate Featurevisor or any other solutions.

will take the learnings from here and build the provider at some point.

thanks everyone! 🙌

